PR #10, opened 2025-02-14
Add support for Clickhouse

mdeeks:
Clickhouse can now be optionally enabled by setting `enable_clickhouse=true`. Similar to our Cloudformation stack, it is a single EC2 instance. 

**I no longer create an NLB or Autoscaling group for this like in Cloudformation.**. Instead the lambdas will point directly at the EC2 instance. If we ever change or replace the EC2 host it will automatically update all references to use the new IP. The NLB in this case wasn't adding anything of value.

I believe I have a way to preserve the Clickhouse metadata EBS volume between replacements too. This will allow us to change things like the AMI or other params and recreate a new Clickhouse instance with a short downtime but without losing any data.

The variables `use_external_clickhouse_address` in the root module, `external_clickhouse_s3_bucket_name`, and `clickhouse_instance_count` in the submodule are for a specific use case for one customer and shouldn't be used by anyone else.

Other Changes:
* Allow passing `custom_domain` and `custom_certificate_arn` for Cloudfront in the root module
* Randomize the RDS final snapshot suffix. (This is to support create/destroy/create/destroy CI workflows)


ankrgyl (approved):
looks solid to me

PR #12, opened 2025-02-20
Add VPC peering modules

mdeeks:
Add two modules to assist with setting up VPC peering. Unfortunately setting up peering in terraform is truly awful and requires a three step apply and coordinating steps with the destination VPC owner.

The Braintrust Data Plane should use the vpc-peering-requester modules to initiate a peering request. The destination VPC should use the vpc-peering-accepter module to accept it. Both modules will set up proper routes to allow bi-directional communication. The README has the details.

These modules do not try to modify security group rules. That should be handled individually depending on the use-case.

Other Changes:
* Stop using launch templates for clickhouse. Unfortunately terraform has a known long-standing bug and will always show drift of the user_data script
* Fix to the deployment_name to allow for numbers and to restrict the length

### DIFF
--- a/.tflint.hcl
+++ b/.tflint.hcl
@@ -5,6 +5,6 @@ plugin "terraform" {
 
 plugin "aws" {
     enabled = true
-    version = "0.36.0"
+    version = "0.37.0"
     source  = "github.com/terraform-linters/tflint-ruleset-aws"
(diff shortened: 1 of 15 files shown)


ankrgyl (approved):
rubber

Issue #35, opened 2025-03-25
Apply fails in us-east-2 due to missing SLR


guseggert:
In `us-east-2` when I first use this module, I get the following error:

```
│ Error: creating KMS Key: operation error KMS: CreateKey, https response error StatusCode: 400, RequestID: f569ff2f-057f-4f9d-82b3-7e46bc363b7e, MalformedPolicyDocumentException: Policy contains a statement with one or more invalid principals.
│ 
│   with module.braintrust-data-plane.module.kms[0].aws_kms_key.braintrust,
│   on .terraform/modules/braintrust-data-plane/modules/kms/main.tf line 1, in resource "aws_kms_key" "braintrust":
│    1: resource "aws_kms_key" "braintrust" {
│ 
╵
```

It looks like the SLR for autoscaling doesn't exist in us-east-2, I was able to get past this error by running:

```
aws iam create-service-linked-role --aws-service-name autoscaling.amazonaws.com --region us-east-2
```

mdeeks:
Thanks @guseggert. Unfortunately it turns out that Terraform does not have a clean way of handling service-linked roles. I'll have to include this step to create the service-linked role as part of the README or possibly a script.

Service-linked roles are created automatically when you interact with parts of the console or create certain resources. In a fresh AWS account it won't exist. Terraform does not have a way to "create resource if it doesn't exist and also ignore it if it already does". 

Scenarios:
1. Service-linked role already exists in the account
Terraform will error saying the role already exists and the plan/apply will fail:
```
Error: creating IAM Service Linked Role (autoscaling.amazonaws.com): operation error IAM: CreateServiceLinkedRole, https response error StatusCode: 400, RequestID: b1178400-d44e-478b-b135-45544b2bf9c2, InvalidInput: Service role name AWSServiceRoleForAutoScaling has been taken in this account, please try a different suffix.
```

2. Service-linked role does not exist.
Terraform will successfully create the role and track it in it's state. Unfortunately that means it will also attempt to destroy the role if you ever want to remove a Braintrust data plane from your account. This will potentially cause an outage for any other service that relies on it.

PR #36, opened 2025-03-26
Add script for service-linked roles

mdeeks:
Related to #35 

Fresh AWS accounts will be missing a service-linked role for autoscaling. There isn't a good way in terraform to create the role and then also not track it in the state (see ticket for more details).

For now we'll just have to let users know they need to run a script once against their AWS accounts. Right now only there is only a role for autoscaling but we likely will have more in the future.


choochootrain (approved):
jeff bozos would approve

PR #49, opened 2025-05-02
Add pg_partman to list of loaded extensions.

manugoyal:
Used for the logs2 migration


mdeeks (approved):

manugoyal:
Pretty sure we actually won't need this. Because `pg_partman` can be dynamically loaded without putting it into shared_preload_libraries.

PR #74, opened 2025-06-26
Enable row references and force Brainstore as default

mdeeks:
This instructs the API to write object store references into the database instead of full payloads. It greatly reduces the amount of data being inserted into postgres and improves performance. This feature requires Brainstore and so BRAINSTORE_DEFAULT is now set to `force`.

The only users who might be negatively impacted by this are ones who are still using Clickhouse or not using Brainstore at all (PG only). There aren't any known Terraform Clickhouse users anymore and PG only mode is no longer supported (much of the UI refuses to operate in this mode already)


manugoyal (approved):
lgtm! Should we add to the changelog in the internal repo?

PR #77, opened 2025-07-09
Use individual SGs instead of Default SG

jeffmccollum:
When deploying the Braintrust data plane, we will no longer use the default VPC security group, but instead create individual security groups for Brainstore, elasticache, rds, and lambdas. 

Existing deployments will have all resources deployed from Terraform switch from using the default security groups to the individual security groups. The quarantine lambdas that are created outside of Terraform will still have the default VPC Security group of the Quarantine VPC, however these can be changed to use the new Quarantine Security Group instead via the AWS Console or via script.


outputs.tf
mdeeks, line 1: We'll want to output all of the new security group IDs here in the root module

modules/database/variables.tf
```diff
@@ -51,9 +51,25 @@ variable "database_subnet_ids" {
...
+variable "lambda_security_group_id" {
+  description = "Security Group ID for the Lambda functions."
+  type        = string
+}
+
+variable "remote_support_security_group_id" {
+  type        = string
+  description = "Security Group ID for the Remote Support bastion host."
+  default     = null
 }
```
mdeeks, line 73: I think we might want to change all of these explicit variables to instead be a single variable that is a list of allowed security groups. Something like `authorized_security_group_ids` maybe. 

We have customers who use these submodules directly so ideally they aren't so coupled with our specific implementations (e.g. they may not use lambdas or have brainstore on ec2)

outputs.tf
jeffmccollum: done!

mdeeks (approved):

PR #95, opened 2025-08-09
Add `monitoring_telemetry` to control enabled telemetry

ibolmo:
Adds a new `monitoring_telemetry` variable which allows for controlling what telemetry is passed back to the Braintrust Control Plane.

- By default it is `status,metrics` which sends basic health telemetry to Braintrust. This telemetry contains no PII or sensitive information. It is highly recommended to leave these enabled as it enables Braintrust to support your self-hosted data plane much more effectively.
- To enable billing telemetry, add `usage` to the `monitoring_telemetry` comma delimited list (`usage,status,metrics`)

Deprecates the previous `enable_billing_telemetry` variable in favor of `monitoring_telemetry`.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -389,9 +389,27 @@ variable "brainstore_vacuum_all_objects" {
 }
 
 variable "monitoring_telemetry" {
-  description = "The telemetry to send to Braintrust's control plane to monitor your deployment. Should be in the form of comma-separated values. Available options are metrics, logs, traces, status, memprof, and usage."
+  description = <<-EOT
+    The telemetry to send to Braintrust's control plane to monitor your deployment. Should be in the form of comma-separated values.
+
+    Available options:
+    - status: Health check information (default)
+    - metrics: System metrics (CPU/memory) and Braintrust-specific metrics like indexing lag (default)
+    - usage: Billing usage telemetry for aggregate usage metrics
+    - memprof: Memory profiling statistics and heap usage patterns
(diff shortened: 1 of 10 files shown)


variables.tf
```diff
@@ -203,10 +203,11 @@ variable "ai_proxy_reserved_concurrent_executions" {
...
 }
 
+# tflint-ignore: terraform_unused_declarations
 variable "enable_billing_telemetry" {
-  description = "Enable billing telemetry. Do not enable this unless instructed by support."
+  description = "DEPRECATED: Billing telemetry is now always enabled. This parameter is ignored."
   type        = bool
-  default     = false
+  default     = true
 }
```
mdeeks, line 211: This should be removed both from here and the submodule. No need to keep it. Even if users for some reason had it set, they will get an error during planning. In that case they would reach out to us or look at the change log and know what to change.

variables.tf
```diff
@@ -394,6 +401,11 @@ variable "brainstore_vacuum_all_objects" {
   default     = false
 }
 
+variable "monitoring_telemetry" {
+  description = "The telemetry to send to Braintrust's control plane to monitor your deployment. Should be in the form of comma-separated values. Available options are metrics, logs, traces, status, memprof, and usage."
```
mdeeks, line 405: We should document here what the various options mean. Specifically we should call out the defaults (status and metrics) do not contain any PII at all.

PR #96, opened 2025-08-13
Remove vacuum all objects config param

manugoyal:
To align with CF template and internal repo change to rely on defaults.


modules/brainstore/templates/user_data.sh.tpl
```diff
@@ -100,7 +100,9 @@ BRAINSTORE_WAL_URI=postgres://$DB_USERNAME:$DB_PASSWORD@${database_host}:${datab
 BRAINSTORE_CACHE_DIR=/mnt/tmp/brainstore
 BRAINSTORE_LICENSE_KEY=${brainstore_license_key}
 BRAINSTORE_DISABLE_OPTIMIZATION_WORKER=${brainstore_disable_optimization_worker}
-BRAINSTORE_VACUUM_OBJECT_ALL=${brainstore_vacuum_all_objects}
+%{ if brainstore_disable_vacuum ~}
```
manugoyal, line 103: I wanted to explicitly turn it off for reader nodes, but rely on the default for writer or reader/writer nodes. So I opted for a flag which only sets it to false if specified. Not sure if there's a better way to do this.

mdeeks: lgtm!

PR #102, opened 2025-08-27
Use BRAINSTORE_READER_ONLY_MODE var to enable reader-only settings

manugoyal:
Aligns with internal-repo PR to use a single env var for enabling "reader-only" functionality.


modules/brainstore/templates/user_data.sh.tpl
```diff
@@ -99,10 +99,8 @@ BRAINSTORE_METADATA_URI=postgres://$DB_USERNAME:$DB_PASSWORD@${database_host}:${
 BRAINSTORE_WAL_URI=postgres://$DB_USERNAME:$DB_PASSWORD@${database_host}:${database_port}/postgres
 BRAINSTORE_CACHE_DIR=/mnt/tmp/brainstore
 BRAINSTORE_LICENSE_KEY=${brainstore_license_key}
+BRAINSTORE_READER_ONLY_MODE=${is_dedicated_reader_node}
 BRAINSTORE_DISABLE_OPTIMIZATION_WORKER=${brainstore_disable_optimization_worker}
```
mdeeks, line 103: Should we also remove this one?

mdeeks (changes requested):

modules/brainstore/templates/user_data.sh.tpl
manugoyal: We could! I left it in because it's still a valid independent configuration option. But if we want to cut down on exposed variables I'm happy to remove it.

PR #121, opened 2025-10-02
Allow for optionally passing in a pre-existing VPC + Remove Clickhouse refs.

mdeeks:
Users can now optionally disable creation of the main VPC that hosts the braintrust services and data stores. An existing VPC and subnets can be passed in instead. By default, the module will continue to create a VPC as it has always.

To disable creation of the main VPC:

```hcl
module "braintrust-data-plane" {
  source = "github.com/braintrustdata/terraform-braintrust-data-plane"
  # ... your existing configuration ...
  # Use existing VPC
  create_vpc = false
  existing_vpc_id                        = "vpc-xxxxxxxxx"
  existing_private_subnet_1_id           = "subnet-xxxxxxxxx"
  existing_private_subnet_2_id           = "subnet-yyyyyyyyy"
  existing_private_subnet_3_id           = "subnet-zzzzzzzzz"
  existing_public_subnet_1_id            = "subnet-aaaaaaaaa"
}
```

There were old references to Clickhouse that were removed. No existing TF users should be on Clickhouse. We dropped support for it earlier in the year. This change removes variables, outputs, and the references to the module that was already disabled.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -8,16 +8,21 @@ module "kms" {
 
 locals {
   kms_key_arn = var.kms_key_arn != "" ? var.kms_key_arn : module.kms[0].key_arn
-  clickhouse_address = var.use_external_clickhouse_address != null ? var.use_external_clickhouse_address : (
-    var.enable_clickhouse ? module.clickhouse[0].clickhouse_instance_private_ip : null
-  )
   bastion_security_group = var.enable_braintrust_support_shell_access ? {
     "Remote Support Bastion" = module.remote_support[0].remote_support_security_group_id
   } : {}
+
+  # VPC configuration - handle both created and existing VPCs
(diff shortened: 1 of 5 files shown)


main.tf
```diff
@@ -222,10 +214,17 @@ module "brainstore" {
...
 
   kms_key_arn = local.kms_key_arn
 }
+
+# Handle state migration since the VPC module became conditional
+# This ensures existing users don't destroy/create their VPC resources when upgrading
+moved {
+  from = module.main_vpc
+  to   = module.main_vpc[0]
+}
```
mdeeks, line 230: This block will have to stay in our TF essentially forever. It instructs TF where to look for the new state and automatically does a move during apply. Since main_vpc is now conditional it will always have a [0] index as its identifier.

KevinGreen (commented):
LGTM, just looks like docs need to be regenerated

jeffmccollum (approved):

PR #127, opened 2025-10-07
Upgrade to AWS provider v6

mdeeks:
Version 6 was released around six months ago and it includes some improvements around multi-region support. 

Existing users might have to run `terraform init -upgrade` to update their terraform lock file to point at the new version. 

Changes:
* The `.name` attribute was renamed to `.region` in the aws_region data source

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ braintrust-sandbox/*
 .idea
 
 scripts/logs*
+.vscode/*


mdeeks:
Closing this since there is no easy rollback from v6 to v5 in case something goes wrong. We'll have to walk customers through this upgrade when it happens.

PR #131, opened 2025-10-11
DRAFT: Allow the use of existing EKS cluster

mdeeks:
Many customers have expressed desire to use their own EKS cluster while also using our Terraform module to set up the base infrastructure (Database, IAM, Security Groups, etc) required by our data plane.

To enable that, major restructuring out of Terraform module was required. Given the size of the change and the risk, this will be marked as a Major release (v2 -> v3).

The core change was to make the Lambdas, Brainstore EC2, and Ingress (Cloudfront, API Gateway) optional. Additionally I needed to extract out the shared pieces that an EKS cluster would need (IAM, Security Groups, S3 buckets).

Unfortunately our existing `services` and `brainstore` modules contained Lambda/EC2 specific resources mixed in with shared resources. I split the two old modules out into five new modules and surgically moved resources (and state) to the new locations:
* Ingress
* Storage
* Services-Common
* API-Lambdas
* Brainstore-EC2

CleanShot 2025-10-10 at 17 21 54@2x

To enable EKS mode these vars should be set:
```
use_deployment_mode_external_eks = true
enable_eks_pod_identity = true

# Optional for further restricting the IAM role security
existing_eks_cluster_arn = "<arn>"
eks_namespace = "<namespace>"
```

Caveats:
* I was unable to make this change in a way that would show no changes for existing customers who use lambda+ec2. The changes are backwards compatible, but will potentially confuse people
  * The IAM Policy for the API Handler had to be split into two policies. One that has Lambda specific policies and another that has generic policies (like S3 bucket permissions)
  * The security group for the API is named `lambda`. There is no way to rename a security group without downtime. Instead I created a new duplicative security group called `api`. Both `lambda` and `api` SGs will be attached to the Lambdas and both will be authorized by the DB, Redis, Brainstore to reach them on their respective ports. In the far future we can probably remove `lambda` after we're sure customers have moved.
* The Brainstore IAM role still has the word `-ec2` in it. This is incorrect but changing the name requires recreating the role which leads to downtime. In EKS deployments the Brainstore IAM role will be named `brainstore-ec2-role`

Todo:
* Consider how a rollback will work.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -64,16 +64,17 @@ module "quarantine_vpc" {
 }
 
 module "database" {
-  source                    = "./modules/database"
-  deployment_name           = var.deployment_name
-  postgres_instance_type    = var.postgres_instance_type
-  multi_az                  = var.postgres_multi_az
-  postgres_storage_size     = var.postgres_storage_size
-  postgres_max_storage_size = var.postgres_max_storage_size
-  postgres_storage_type     = var.postgres_storage_type
-  postgres_version          = var.postgres_version
-  database_subnet_ids       = local.database_subnet_ids
-  vpc_id                    = local.main_vpc_id
+  source                              = "./modules/database"
+  deployment_name                     = var.deployment_name
+  postgres_instance_type              = var.postgres_instance_type
+  multi_az                            = var.postgres_multi_az
+  postgres_storage_size               = var.postgres_storage_size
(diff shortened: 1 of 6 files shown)


mdeeks:
Closing this in favor of https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/141 which removes aws v6 upgrade and a few other risky steps

PR #139, opened 2025-10-16
Fixes to validation for RDS gp3 throughput/IOPS restrictions

mdeeks:
RDS has confusing restrictions for gp3 volumes based on size. This adds validation logic so users can be notified of it during plan time rather than apply time.

You can't specify IOPS/Throughput if the volume is less than 400GB. And if it is >=400GB it must meet minimums.

CleanShot 2025-10-16 at 13 28 53@2x
From here: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ braintrust-sandbox/*
 .idea
 
 scripts/logs*
+.vscode


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "window.title": "🌱 terraform-aws-braintrust-data-plane/validate-db-params"
```
KevinGreen: Looked like this snuck in

KevinGreen (approved):

PR #173, opened 2025-12-18
valkey migration

jgoulah:


modules/elasticache/main.tf
```diff
@@ -2,6 +2,12 @@ locals {
   common_tags = merge({
     BraintrustDeploymentName = var.deployment_name
   }, var.custom_tags)
+
+  # Migration logic:
+  # "default" = both old and new exist (migration state)
+  # "serverless" = only new exists (final state)
+  use_old_cluster = var.cache_mode == "default"
+  use_serverless  = true  # Always create serverless during migration
```
mdeeks: This shouldn't be needed actually. In all cases serverless will exist. So no need for a count.

modules/elasticache/outputs.tf
```diff
@@ -1,15 +1,27 @@
 output "redis_endpoint" {
-  value       = aws_elasticache_cluster.main.cache_nodes[0].address
+  value = local.use_serverless ? (
```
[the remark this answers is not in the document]
jgoulah: yup, the other change kinda implies that too, thanks!

modules/elasticache/main.tf
```diff
@@ -20,6 +27,29 @@ resource "aws_elasticache_cluster" "main" {
...
+  }
+}
+
+resource "aws_elasticache_serverless_cache" "valkey" {
+  engine = "valkey"
+  name   = "${var.deployment_name}-valkey"
+
+  cache_usage_limits {
+    data_storage {
+      maximum = 10
```
mdeeks: These should be exposed as vars

modules/elasticache/main.tf
```diff
@@ -20,6 +27,29 @@ resource "aws_elasticache_cluster" "main" {
...
+      unit    = "GB"
+    }
+    ecpu_per_second {
+      maximum = 5000
+    }
+  }
+
+  security_group_ids = [aws_security_group.elasticache.id]
+  subnet_ids         = var.subnet_ids
+  tags               = local.common_tags
```
mdeeks: Should use a kms key

mdeeks: Should set and expose the major engine version too. We may want to lock onto specific versions.

modules/elasticache/outputs.tf
```diff
@@ -1,15 +1,15 @@
 output "redis_endpoint" {
```
mdeeks: We should consider renaming these since it is no longer redis

PR #177, opened 2025-12-19
Add CloudWatch Log Groups with Retention for Lambda Functions

jgoulah:
Creates explicit CloudWatch log groups with 90-day retention policies for all Lambda functions to prevent indefinite log storage and reduce costs.

Problem
Lambda functions currently reference log groups in their logging_config, but these log groups aren't explicitly created with retention policies. AWS automatically creates them with indefinite retention, causing logs to accumulate forever and unnecessarily increase storage costs.

Solution
Added aws_cloudwatch_log_group resources for each Lambda function with:
- 90-day retention period - balances compliance/debugging needs with cost management
- KMS encryption - consistent with other infrastructure encryption standards
- Lifecycle management - logs automatically expire after retention period

Impact
- Cost optimization: Automated cleanup of old logs prevents unbounded storage growth
- Security: Log data encrypted with customer-managed KMS key
- Operational: 90 days provides sufficient window for troubleshooting and compliance
- No breaking changes: Log groups created before Lambda functions deploy

### DIFF
--- a/modules/kms/main.tf
+++ b/modules/kms/main.tf
@@ -49,6 +49,27 @@ resource "aws_kms_key" "braintrust" {
           Resource = "*"
         },
         {
+          Sid    = "Allow CloudWatch Logs to use the key"
+          Effect = "Allow"
+          Principal = {
+            Service = "logs.${data.aws_region.current.id}.amazonaws.com"
+          }
+          Action = [
+            "kms:Encrypt",
+            "kms:Decrypt",
+            "kms:ReEncrypt*",
+            "kms:GenerateDataKey*",
+            "kms:DescribeKey",
+            "kms:CreateGrant"
+          ]
+          Resource = "*"
+          Condition = {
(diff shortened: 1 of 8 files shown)


jgoulah:
90 days is arbitrary but I'm guessing we dont want _infinite_ as it stands now

jgoulah:
it was [noted](https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/177#discussion_r2635600271) these logs would fail because the KMS key policy doesn't include CloudWatch Logs service.  I wasn't sure if we require this to be encrypted or not, if so we could add the CloudWatch Logs service principal to the KMS key policy.  For now I have simply removed the encryption.

mdeeks:
Closing until we have time later to test this. The main issue is that the groups will already exist since they get dynamically created. TF will fail because it exists. This requires some testing to make sure things work as expected. Will repoen later

PR #238, opened 2026-04-30
Add API ECS for Brainstore or Private only dataplane

jeffmccollum:


modules/api-ecs/main.tf
```diff
@@ -0,0 +1,383 @@
...
+    TELEMETRY_DISABLE_AGGREGATION      = tostring(var.disable_billing_telemetry_aggregation)
+    TELEMETRY_LOG_LEVEL                = var.billing_telemetry_log_level
+    INSERT_LOGS2                       = "true"
+    NODE_MEMORY_PERCENT                = "80"
+    ALLOW_CODE_FUNCTION_EXECUTION      = tostring(var.code_function_execution_mode == "api_ecs")
+    AI_PROXY_FN_URL                    = "http://127.0.0.1:8000"
+    },
+    local.using_brainstore_fast_reader ? {
+      BRAINSTORE_FAST_READER_URL           = "http://${var.brainstore_fast_reader_hostname}:${var.brainstore_port}"
+      BRAINSTORE_FAST_READER_QUERY_SOURCES = "summaryPaginatedObjectViewer [realtime],summaryPaginatedObjectViewer,a602c972-1843-4ee1-b6bc-d3c1075cd7e7,traceQueryFn-id,traceQueryFn-rootSpanId,fullSpanQueryFn-root_span_id,fullSpanQueryFn-id"
```
mdeeks, line 36: This shouldn't be duplicated inline like this. It needs to be extracted out somewhere that both ECS and Lambda can re-use it

variables.tf
```diff
@@ -456,6 +456,204 @@ variable "gateway_braintrust_app_url" {
...
+  validation {
+    condition     = var.api_ecs_min_count >= 1
+    error_message = "api_ecs_min_count must be at least 1."
+  }
+}
+
+variable "api_ecs_max_count" {
+  description = "Maximum number of API ECS tasks."
+  type        = number
+  default     = 16
```
mdeeks, line 523: We should default this much higher. Maybe 64

variables.tf
```diff
@@ -456,6 +456,204 @@ variable "gateway_braintrust_app_url" {
...
+  validation {
+    condition     = !(var.use_api_ecs_for_brainstore_ai_proxy_url && var.use_deployment_mode_external_eks)
+    error_message = "use_api_ecs_for_brainstore_ai_proxy_url cannot be true when use_deployment_mode_external_eks is true."
+  }
+}
+
+variable "api_ecs_cpu" {
+  description = "CPU units for the API ECS task definition."
+  type        = number
+  default     = 2048
```
mdeeks, line 500: Is our API container setup to do multiprocessing yet? if not we should not set this to 2. It should be one since it is a single node process.

variables.tf
```diff
@@ -456,6 +456,204 @@ variable "gateway_braintrust_app_url" {
...
+  validation {
+    condition     = var.api_ecs_max_count >= var.api_ecs_min_count
+    error_message = "api_ecs_max_count must be greater than or equal to api_ecs_min_count."
+  }
+}
+
+variable "api_ecs_cpu_target_value" {
+  description = "Target average CPU utilization percentage for API ECS autoscaling."
+  type        = number
+  default     = 40
```
mdeeks, line 534: This is dependent on the CPU count and if we're doing multiprocessing or not right?

variables.tf
```diff
@@ -456,6 +456,204 @@ variable "gateway_braintrust_app_url" {
...
+  validation {
+    condition = contains([
+      1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180,
+      365, 400, 545, 731, 1096, 1827, 2192, 2557, 2922, 3288, 3653
+    ], var.api_ecs_log_retention_days)
+    error_message = "api_ecs_log_retention_days must be a valid CloudWatch Logs retention value."
+  }
+}
+
+variable "api_ecs_target_group_deregistration_delay_seconds" {
```
mdeeks, line 567: I dont think we need to expose this to users. We should pick values that we think is sensible. 300 seems fine

variables.tf
```diff
@@ -456,6 +456,204 @@ variable "gateway_braintrust_app_url" {
...
+  }
+}
+
+variable "api_ecs_extra_env_vars" {
+  description = "Extra environment variables for the API ECS container."
+  type        = map(string)
+  default     = {}
+}
+
+variable "api_ecs_code_function_execution_mode" {
```
mdeeks, line 584: I dont think this should exist? Why wouldn't we always just use lambdas for user code?

modules/brainstore-ec2/templates/user_data.sh.tpl
```diff
@@ -161,17 +161,21 @@ ${env_key}=${env_value}
...
 
-BRAINSTORE_AI_PROXY_URL=$(aws ssm get-parameter \
-  --name /braintrust/${deployment_name}/ai-proxy-url \
-  --query 'Parameter.Value' \
-  --output text \
-  --region ${aws_region})
-
-if [ -n "$BRAINSTORE_AI_PROXY_URL" ]; then
-  echo "BRAINSTORE_AI_PROXY_URL=$BRAINSTORE_AI_PROXY_URL" >> /etc/brainstore.env
+if [ "${use_api_ecs_for_brainstore_ai_proxy_url}" = "true" ] && [ -n "${api_ecs_url}" ]; then
```
mdeeks, line 164: I believe this created a circular dependency for the case when we move to all ECS. On first launch the API ECS container requires Brainstore and Brainstore requires the API ECS container. The reason it was put into SSM is to allow for brainstore to be created but fail to startup until the AI proxy exists

mdeeks:
Closing this in favor of a smaller more targeted PRs that split this into 1) ECS for AI proxy 2) ECS for everything 3) Private data plane 4) adding support for ABAC in S3

PR #242, opened 2026-05-07
Add IAM policy to allow Brainstore to assume export roles with bt:* ExternalId

jeffmccollum:
### Motivation

- Grant the Brainstore IAM role the ability to assume export roles that present an external ID prefixed with `bt:` to support cross-account export workflows.

### Description

- Introduces `aws_iam_role_policy.brainstore_export_assume_role` which allows `sts:AssumeRole` on all resources with a `StringLike` condition `sts:ExternalId = "bt:*"`.
- Attaches the new JSON-encoded policy to the existing `aws_iam_role.brainstore_role` alongside existing S3, SecretsManager, and CloudWatch policies.

### Testing

- No automated tests were run as part of this change.

### DIFF
--- a/modules/services-common/iam-brainstore.tf
+++ b/modules/services-common/iam-brainstore.tf
@@ -126,6 +126,8 @@ resource "aws_iam_role_policy" "brainstore_secrets_access" {
 }
 
 resource "aws_iam_role_policy" "brainstore_export_assume_role" {
+  count = var.brainstore_enable_export ? 1 : 0
+
   name = "BrainstoreExportAssumeRole"
   role = aws_iam_role.brainstore_role.id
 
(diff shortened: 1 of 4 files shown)


modules/services-common/iam-brainstore.tf
```diff
@@ -125,6 +125,29 @@ resource "aws_iam_role_policy" "brainstore_secrets_access" {
   })
 }
 
+resource "aws_iam_role_policy" "brainstore_export_assume_role" {
+  count = var.brainstore_enable_export ? 1 : 0
+
+  name = "BrainstoreExportAssumeRole"
```
KevinGreen:
```suggestion
  name = "export-assume-role"
```
For more consistency with the tf naming

KevinGreen (dismissed):

KevinGreen (approved):

PR #246, opened 2026-05-18
Deploy API as ECS Task for Brainstore AI Proxy

jeffmccollum:
This change allows moving the AI proxy lambda into an ECS task instead. This feature is opt-in right now but will soon be the default going forward as we move to phase out lambdas completely.

This can be enabled by setting `enable_api_ecs = true`.

Background:
Lambas are poorly suited for calls to AI providers primarily due to costs, but also due to various other limitations around max run time, streaming responses, and configurations that allow for private data planes.  

Calls to AI providers are often very long (seconds to minutes) and lambda charges by the second. A single lambda might be simply waiting for network IO and consuming almost no CPU. Meanwhile the costs continue to accumulate while it does almost nothing.

Additionally there is no supported configuration that allows for a fully private data plane. Lambdas can only be fronted by Cloudfront, API Gateway, or ALBs. The first is always public, the second has limitations around streaming responses, and the third has 1MB limits on requests and response sizes.


modules/services-common/iam-api.tf
```diff
@@ -86,56 +96,71 @@ resource "aws_iam_role_policy_attachment" "api_handler_additional_policy" {
...
+          Sid    = "ECSExec"
+          Effect = "Allow"
+          Action = [
+            "ssmmessages:CreateControlChannel",
+            "ssmmessages:CreateDataChannel",
+            "ssmmessages:OpenControlChannel",
+            "ssmmessages:OpenDataChannel",
+          ]
+          Resource = "*"
+        }
```
mdeeks, line 161: Lets remove ecs exec until we can expose it in a way that customers will be happy with.

mdeeks: Actually, never mind. It doesn't seem possible to further restrict things here in this policy. This just allows a role with proper permissions to ecs exec in. It delegates permissions to the customer account.

README.md
```diff
@@ -69,11 +78,25 @@ See the [`examples/cloudfront-logging`](examples/cloudfront-logging) directory f
 
 ## Advanced: Customized Deployments
 
+### API ECS Autoscaling
```
mdeeks: I dont think we need this in the docs?

jeffmccollum: removed.

.github/workflows/create-release.yml
```diff
@@ -8,7 +8,7 @@ on:
         required: true
         type: string
       services_version:
-        description: 'Lock onto Braintrust services version (e.g., v1.2.3). Optional.'
+        description: 'Lock onto Braintrust services/API ECS version (e.g., v1.2.3). Optional.'
```
mdeeks: Remove the API ECS part. They are, as a whole, known as "Braintrust Services".

jeffmccollum: 👍

modules/api-ecs/main.tf
```diff
@@ -0,0 +1,325 @@
...
+  kms_key_id        = var.kms_key_arn
+
+  tags = merge({
+    Name = "${var.deployment_name}-api-ecs-logs"
+  }, local.common_tags)
+}
+
+resource "aws_security_group" "alb" {
+  name        = "${var.deployment_name}-api-ecs-alb"
+  description = "Security group for private API ECS ALB"
```
mdeeks: remove "private"

modules/api-ecs/main.tf
```diff
@@ -0,0 +1,325 @@
...
+  }
+}
+
+resource "aws_ecs_service" "api_ecs" {
+  name                              = "${var.deployment_name}-api-ecs"
+  cluster                           = var.ecs_cluster_arn
+  task_definition                   = aws_ecs_task_definition.api_ecs.arn
+  desired_count                     = var.min_count
+  launch_type                       = "FARGATE"
+  force_new_deployment              = true
```
mdeeks, line 289: This should be false. Otherwise it will always redeploy when running terraform apply

modules/api-ecs/outputs.tf
```diff
@@ -0,0 +1,39 @@
...
+
+output "http_url" {
+  description = "HTTP URL for API ECS ALB."
+  value       = local.api_ecs_url
+}
+
+output "effective_url" {
+  description = "Preferred URL for API ECS ALB."
+  value       = local.api_ecs_url
+}
```
mdeeks: I think this is extra?

modules/api-ecs/main.tf
jeffmccollum: I don't believe so, force_new_deployment doesn't mean every Terraform apply trigger a new deployment. 
We could do that if we did something like this
https://registry.terraform.io/providers/-/aws/6.45.0/docs/resources/ecs_service#redeploy-service-on-every-apply

modules/api-ecs/outputs.tf
jeffmccollum: Yep left over

modules/api-ecs/main.tf
mdeeks: No, that is what it does. What you linked to is an alternate way to do it. The docs
> [force_new_deployment](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_service#force_new_deployment-1) - (Optional) Enable to force a new task deployment of the service. This can be used to update tasks to use a newer Docker image with same image/tag combination (e.g., myimage:latest), roll Fargate tasks onto a newer platform version, or immediately deploy ordered_placement_strategy and placement_constraints updates.

mdeeks: Huh weird. You're right. This var is so strange. It doesn't actually let you force a new deployment except in certain circumstances. Okay lets leave it.

main.tf
```diff
@@ -345,6 +421,7 @@ module "brainstore" {
   fast_reader_instance_type             = var.brainstore_fast_reader_instance_type
   extra_env_vars_fast_reader            = var.brainstore_extra_env_vars_fast_reader
   cache_file_size_fast_reader           = var.brainstore_cache_file_size_fast_reader
+  ai_proxy_url_ssm_parameter_name       = local.brainstore_ai_proxy_url_ssm_parameter_name
```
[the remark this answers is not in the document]
mdeeks: This is by design. It is a known cyclical dependency that is only really an issue during initial deployment of the stack. During initial deploy brainstore nodes may start and fail until the API container is running.

variables.tf
```diff
@@ -456,6 +456,127 @@ variable "gateway_braintrust_app_url" {
...
+  validation {
+    condition     = var.api_ecs_version_override == null || var.api_ecs_version_override != ""
+    error_message = "api_ecs_version_override must be null or a non-empty string."
+  }
+}
+
+variable "enable_api_ecs" {
+  type        = bool
+  description = "Create the dedicated internal API ECS service."
+  default     = false
```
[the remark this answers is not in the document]
mdeeks: I don't think we need an example yet. Let's wait until we roll out ECS fully.

mdeeks (approved):

PR #249, opened 2026-05-20
Add org ID allowlists and URL-security controls

ibolmo:
## Description

- Add `allowed_org_ids` to the root module and API service modules, validate it as a comma-separated list of Braintrust Org UUIDs, and pass it to Lambda/ECS API services as `ALLOWED_ORG_IDS`.
- Require `primary_org_name` when `braintrust_org_name` is empty or `"*"`, so service-token management still has a primary org boundary for multi-org or ID-only deployments.
- Add URL-security controls for Braintrust backends: `unsafe_url_request_mode`, `url_security_dns_servers`, and `url_security_allow_cidrs`.
- Inject non-empty URL-security settings into the Lambda API handler, ECS API, and AI gateway environment variables.
- Update README and examples to show preferred org-ID configuration and the customer-facing URL-security inputs.

## Context

Hybrid data planes need stable organization-ID authorization instead of relying only on organization names. This keeps existing name-based deployments compatible while letting new deployments configure ID-based access explicitly.

This also brings Terraform parity for runtime URL-security controls used when Braintrust services make outbound requests to externally supplied URLs. Leaving these inputs empty preserves the application defaults.

## Testing

- `terraform validate`


KevinGreen (dismissed):

modules/brainstore-ec2/VERSIONS.json
```diff
@@ -1,3 +1,3 @@
 {
     "brainstore": "v2.4.0"
```
soldatchenko: There was a recent release for this module that bumped this to 2.5.0

modules/services/VERSIONS.json
```diff
@@ -1,3 +1,3 @@
 {
     "lambda_version_tag": "v2.4.0"
```
soldatchenko: There was a recent release for this module that bumped this to 2.5.0

modules/api-ecs/VERSIONS.json
```diff
@@ -1,3 +1,3 @@
 {
     "api": "v2.4.0"
```
soldatchenko: There was a recent release for this module that bumped this to 2.5.0

soldatchenko (changes requested):

variables.tf
```diff
@@ -731,6 +754,17 @@ variable "outbound_rate_limit_window_minutes" {
   default     = 1
 }
 
+variable "unsafe_url_request_mode" {
+  description = "Controls how Braintrust backends handle outbound requests to user-supplied URLs that fail URL-security checks, such as URLs resolving to private or reserved IP ranges. Use off to allow, proxy to proxy, warn to allow with warnings, or reject to block. Leave empty to use the application default of warn."
+  type        = string
+  default     = ""
```
soldatchenko, line 760: Do we want the module default to be explicit `warn` instead of empty? Empty will preserve the app default behavior, but makes the security posture depend on image version (which could be okay, but wanted to ask)

soldatchenko (changes requested):

variables.tf
ibolmo: i'd prefer empty because that allows the api handler (etc) to default to whatever we choose

soldatchenko (approved):

PR #250, opened 2026-05-21
[internal] Add datadog agent in the ECS API if observability key is provided

mdeeks:
This enables datadog monitoring in ECS API if provided. Generally this is only for internal Braintrust use, but some customers use it as well. It is not recommended for customers to set this in their data plane.

When the API key is provided it is now stored as an AWS secrets manager secret and referenced in the service.


jeffmccollum (approved):

mdeeks:
This was done in another PR

PR #258, opened 2026-06-01
Add sa-east-1 to services region allowlist

mdeeks:
Add sa-east-1 to supported regions. Only v2.1.1 of the data plane and higher are supported in this region.


btaaron (approved):
Lgtm ✅

PR #270, opened 2026-06-17
Add optional ExternalId to BraintrustManagementRole trust policy

kati-k:
## Description 
Generate and add mandatory ExternalId to the trust policy (management-role-trust-policy.json) in create-management-role.sh script. The BYOC customers need to share this new ExternalId with Braintrust.

Conceptually the biggest change is that when the ExternalId is used the trust policy is now written to the update-assume-role-policy call as a parameter and not included as a file. 

## Context 
A BYOC customer has an internal requirement for all 3rd parties using ExternalId when assuming role on AWS. It is not a bad habit in general and we decided to make it mandatory for all new BYOC customers.


managed-byoc/create-management-role.sh
```diff
@@ -8,21 +8,24 @@ ROLE_POLICY_PATH="$SCRIPT_DIR/policies/management-role-policy.json"
...
 PROFILE="${AWS_PROFILE:-default}"
 INLINE_POLICY_NAME="${INLINE_POLICY_NAME:-BraintrustManagementRolePolicy}"
+EXTERNAL_ID="${EXTERNAL_ID:-}"
 MAX_SESSION_DURATION_SECONDS=14400
 
 usage() {
   cat <<'EOF'
 Usage:
-  bash managed-byoc/create-management-role.sh [--profile PROFILE] [--role-name ROLE_NAME]
+  bash managed-byoc/create-management-role.sh [--profile PROFILE] [--role-name ROLE_NAME] [--external-id EXTERNAL_ID]
```
mdeeks: Lets make this auto generated as a UUID. External ID are only safe if they are not guessable. If people provide this it has a higher chance to be guessable. How about we make this "braintrust-<UUID>"? Don't let the user provide a value. Also we should make this the default and output the value so they can share it back with us.

managed-byoc/create-management-role.sh
```diff
@@ -34,6 +35,41 @@ require_cmd() {
...
+}
+
+resolve_external_id() {
+  if aws iam get-role --profile "$PROFILE" --role-name "$ROLE_NAME" >/dev/null 2>&1; then
+    local existing
+    existing="$(aws iam get-role \
+      --profile "$PROFILE" \
+      --role-name "$ROLE_NAME" \
+      --query 'Role.AssumeRolePolicyDocument' \
+      --output text | jq -r '.Statement[0].Condition.StringEquals."sts:ExternalId" // empty')"
```
soldatchenko: I think this should use `--output json` instead of `--output text` before piping into jq. The AWS CLI text output can flatten structured objects, so this may not always be valid JSON for jq, especially on the existing-role path where we’re trying to preserve the current External ID

soldatchenko:
One operational thing I’d call out in the README/runbook: if this script is re-run against an existing role that does not already have an External ID, it will add one, and Braintrust won’t be able to assume the role until we store/wire that External ID on our side

soldatchenko (changes requested):

managed-byoc/create-management-role.sh
kati-k: Done, and moved all json parsing to jq. In other words, removed --query line and moved the handling to jq.

soldatchenko (approved):
LGTM!

PR #276, opened 2026-06-29
[INF-173] Move gateway ALB to services-common and auto-wire GATEWAY_URL

erikdw:
## Summary

Part of fixing: [INF-173](https://linear.app/braintrustdata/issue/INF-173)

- Move the internal gateway ALB (SG, listener, target group) from `gateway-ecs` into `services-common` so the ALB DNS is available before gateway ECS wiring (breaks the dependency cycle for auto-wiring `GATEWAY_URL`).
- Split `enable_ai_gateway` into **`create_ai_gateway`** (infra: ALB + gateway ECS) and **`enable_ai_gateway`** (wire `GATEWAY_URL` on APIHandler, AIProxy, and ECS API). Supports a two-step prod cutover: create infra first, flip wiring after health checks.
- When `enable_ai_gateway = true`, set `GATEWAY_URL` on **APIHandler**, **AIProxy**, and **ECS API** in a single `terraform apply` — no manual `internal_gateway_url` / two-phase hack.
- **No `moved` blocks** for gateway ALB resources. Existing stacks that already applied the old layout will **destroy/recreate** the gateway ALB (new DNS). Acceptable because private gateway is not in broad production use yet; gateway ECS tasks may recycle when the target group is replaced.
- Do **not** merge `GATEWAY_URL` into `MigrateDatabaseFunction` or crons — avoids publishing new Lambda versions and re-running migrations on upgrade.
- Add `terraform_data.gateway_http_listener` in `gateway-ecs` to enforce listener dependency ordering across modules (see below).
- Add root `gateway_url` output and upgrade guidance in `AGENTS.md`.

**Out of scope for this PR** (follow-up): CloudFront `/v1/proxy` VPC origin cutover to private gateway ALB (bypass AI Proxy Lambda), AI Proxy Lambda removal.

### `terraform_data.gateway_http_listener` (why it exists)

Before this PR, the gateway HTTP listener lived in the same `gateway-ecs` module as the ECS service, so Terraform could order “create listener → attach ECS service” within one module.

After moving the listener to `services-common`, `gateway-ecs` only receives `gateway_http_listener_arn` as an input. Passing an output into a variable does **not** always make Terraform wait for that listener to exist before updating the ECS service — especially on upgrades where the listener is being destroyed and recreated in another module.

`terraform_data` is a small state-only resource that records the listener ARN and gives `aws_ecs_service.gateway` an explicit `depends_on`. That forces the ECS service update to run **after** the new listener in `services-common` exists. It creates no AWS resources (plan shows `+ create` only in Terraform state); it is dependency glue, not new infrastructure.

## Test plan

- [x] Greenfield apply on `erikdw-sandbox` (`create_ai_gateway=true`, `enable_ai_gateway=true`, `use_global_ai_gateway_origin=false`, `BRAINSTORE_ASYNC_SCORING_USE_GATEWAY=all`) — ~175 resources, gateway ECS healthy
- [x] `terraform output gateway_url` returns internal ALB URL; `GATEWAY_URL` present on APIHandler / ECS API when `enable_ai_gateway=true`, absent on MigrateDatabaseFunction
- [x] In-VPC async scoring path: brainstore → internal API ECS ALB → api-ts `forwardToGateway` → internal gateway ALB → OpenAI (gateway CloudWatch logs show `request.source=async_scoring_*`, `passthrough_endpoint=false`)
- [x] In-place upgrade on a deployment with gateway ALB already in `gateway_ecs` state (plan shows ALB destroy/recreate in `services_common`, no `moved` blocks)
- [x] Two-step cutover: `create_ai_gateway=true` + `enable_ai_gateway=false` leaves caller-supplied `GATEWAY_URL` unchanged; flip `enable_ai_gateway=true` wires internal ALB URL
- [x] `terraform plan` with `create_ai_gateway=false` — no gateway resources created

### DIFF
--- a/modules/gateway-ecs/main.tf
+++ b/modules/gateway-ecs/main.tf
@@ -190,37 +190,6 @@ resource "aws_cloudwatch_log_group" "service" {
   }, local.common_tags)
 }
 
-resource "aws_security_group" "alb" {
-  name        = "${var.deployment_name}-gateway-alb"
-  description = "Security group for private gateway ALB"
-  vpc_id      = var.vpc_id
-  tags = merge({
-    Name = "${var.deployment_name}-gateway-alb"
-  }, local.common_tags)
-}
-
-resource "aws_security_group_rule" "alb_ingress_from_authorized_security_groups" {
-  for_each = var.authorized_security_groups
-
-  type                     = "ingress"
-  from_port                = 80
-  to_port                  = 80
-  protocol                 = "tcp"
-  source_security_group_id = each.value
-  description              = "Allow HTTP traffic from ${each.key}."
-  security_group_id        = aws_security_group.alb.id
-}
-
-resource "aws_security_group_rule" "alb_egress_all" {
-  type              = "egress"
-  from_port         = 0
-  to_port           = 0
-  protocol          = "-1"
-  cidr_blocks       = ["0.0.0.0/0"]
-  description       = "Allow all outbound traffic from gateway ALB"
-  security_group_id = aws_security_group.alb.id
-}
-
 resource "aws_security_group" "task" {
   name        = "${var.deployment_name}-gateway-task"
   description = "Security group for gateway ECS tasks"
@@ -235,7 +204,7 @@ resource "aws_security_group_rule" "task_ingress_from_alb" {
   from_port                = local.container_port
   to_port                  = local.container_port
   protocol                 = "tcp"
-  source_security_group_id = aws_security_group.alb.id
+  source_security_group_id = var.alb_security_group_id
   description              = "Allow inbound traffic from gateway ALB to gateway tasks"
   security_group_id        = aws_security_group.task.id
 }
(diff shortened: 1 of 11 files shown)


erikdw:
@codex review

modules/gateway-ecs/main.tf
```diff
@@ -403,13 +323,11 @@ resource "aws_ecs_service" "gateway" {
   }
 
   load_balancer {
-    target_group_arn = aws_lb_target_group.gateway.arn
+    target_group_arn = var.target_group_arn
```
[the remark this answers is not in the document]
erikdw: Addressed in aea0d9d. Root module passes gateway_http_listener_arn from services_common into gateway_ecs. terraform_data.gateway_http_listener records that ARN and aws_ecs_service.gateway depends on it, so CreateService runs only after the listener associates the target group with the ALB. Greenfield sandbox apply with enable_ai_gateway=true passed.

- Cursor

variables.tf
```diff
@@ -335,10 +335,21 @@ variable "redis_authorized_security_groups" {
...
 
+variable "create_ai_gateway" {
+  description = "Create the private gateway infrastructure (internal ALB and ECS gateway service)."
+  type        = bool
+  default     = false
+}
+
 variable "enable_ai_gateway" {
-  description = "Enable ECS gateway service deployment (Fargate with private ALB)"
+  description = "Wire GATEWAY_URL on APIHandler, AIProxy, and ECS API to the private gateway. Requires create_ai_gateway."
```
KevinGreen: We should add a warning here to say that the create_ai_gateway should be applied in a separate step before enabling the ai gateway

erikdw: Done here: https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/276/commits/060c4127af78cd47012fbcf25a3a7801eea972ae

KevinGreen (approved):

PR #283, opened 2026-07-09
Fix Brainstore AI proxy SSM dependency

soldatchenko:
Fixes Brainstore startup ordering for deployments that use the AI proxy URL from SSM.

On new deployments, Brainstore can start before the AI proxy URL parameter is available in SSM. When that happens, Brainstore fails to resolve `BRAINSTORE_AI_PROXY_URL`, does not become healthy behind the NLB, and the ASG may replace instances during apply.

Brainstore now receives the SSM parameter name from the Terraform resource that creates it, giving Terraform an explicit dependency before launching Brainstore EC2 instances. The startup script also retries the SSM lookup for a bounded period to tolerate short propagation delays.

- Non-breaking bug fix with possible launch template updates on existing deployments


soldatchenko:
I deployed and tested a data-plane from this branch.

mdeeks (approved):

PR #284, opened 2026-07-10
Add first class audit log attributes for 2.6.0 & eu-west-3

JPurcell-Braintrust:
## Description

Adds first-class Terraform variables for enabling BTQL `query.read` audit logging by Braintrust Org UUID:

- `btql_audit_logs_strict_org_ids`
- `btql_audit_logs_best_effort_org_ids`

These variables are wired through to both Lambda

 This also adds `eu-west-3` to the services Lambda layer supported region list.

## Context

Data plane 2.6.0 introduced self-hosted audit logging support. Administrative audit logs work with the base data plane behavior, but BTQL/data-read `query.read` audit logs are intentionally gated by org-specific environment variables because they can be high volume.

Before this change, customers had to use lower-level extra env var hooks to set the BTQL audit logging org IDs manually. 

## Testing
Ran Lambda setup and tested validations with duplicate values and values in both variables


brianvans (approved):
> This also adds eu-west-3 to the services Lambda layer supported region list.

Thank you 🏆 🙇

PR #286, opened 2026-07-14
add assume role access for gateway

knjiang:
## Description
  

 This PR adds an inline IAM policy to the Gateway ECS task role, `${deployment_name}-gateway-task`, allowing Gateway tasks to assume customer-provided IAM roles.

This role is used to allow for the `AssumeRole` auth method in braintrust -> https://www.braintrust.dev/docs/integrations/ai-providers/bedrock#connect-bedrock-to-braintrust which is provided as a auth method in the gateway/braintrust platform.

  ## Context
  

  Gateway executes with its own ECS task role, so granting this permission to the shared API role would not authorize
  Gateway’s STS calls.

  This follows the existing cross-account S3 export pattern in modules/services-common/iam-api.tf:162, which already
  allows:

```
  Action   = "sts:AssumeRole"
  Resource = "*"
  Condition = {
    StringLike = {
      "sts:ExternalId" = "bt:*"
    }
  }
```

  The new policy applies that same external-ID boundary to the dedicated Gateway task role in modules/gateway-ecs/
  main.tf:286.

  ## Testing
  

  - Ran terraform fmt -check -recursive.
  - Ran terraform validate.
  - Ran git diff --check.

PR #290, opened 2026-07-16
Private gateway ALB module + CloudFront /v1/proxy VPC origin

erikdw:
## Summary
- **ALB move:** Extract the private AI gateway ALB from `services-common` into `modules/gateway-alb` so callers can read `GATEWAY_URL` / ALB DNS without depending on `gateway-ecs` (Terraform cycle break). Temporary `moved` blocks remount existing state (`services_common` → `gateway_alb[0]`) — keep until every gateway stack has applied once, then remove in a follow-up.
- **`/v1/proxy` re-routing:** Add `use_private_ai_gateway_origin` (mutually exclusive with `use_global_ai_gateway_origin`). When enabled, CloudFront sends `/v1/proxy` (and `/v1/proxy/*`) to a **VPC origin** on the private gateway ALB. Includes CloudFront prefix-list ingress on the gateway ALB SG.
- Greenfield-safe: private VPC origin `count` uses the boolean flag only (not ALB ARN/DNS), so `create_vpc=true` plans succeed.

### Out of scope (follow-up)
CloudFront VPC-origin **unsupported AZ filtering** for both `ApiEcsOrigin` (already on `main` via #287) and `PrivateGatewayOrigin`. Same footgun exists today for ECS API; harden both origins together later.

### `/v1/proxy` origin priority (after this change)
1. Global hosted gateway (`use_global_ai_gateway_origin`)
2. **Private gateway VPC origin** (`use_private_ai_gateway_origin`) ← new
3. ECS API ALB (`enable_ecs_api`)
4. AI Proxy (legacy)

## Test plan
- [x] `terraform validate` on module root after rebase onto `main` (#286 / #287)
- [x] Sandbox upgrade: ALB remounted via `moved` without replace; `enable_ai_gateway` + private origin smoke (`Hello world!` on `/v1/proxy`)
- [x] Re-apply minimized branch (no AZ filtering) on a sandbox / greenfield when convenient
- [x] Applied on a production dataplane so `moved` blocks remounted state; follow-up to delete temporary moves after broader rollout
- [ ] Track follow-up: VPC-origin AZ filtering for ApiEcsOrigin + PrivateGatewayOrigin


erikdw:
Addressed Claude review feedback:

1. **High — greenfield `for_each`:** `data.aws_subnet.private` is now gated to `create_vpc = false`. For `create_vpc = true`, ALB subnet filtering uses known AZ name → zone ID locals (`zipmap` of `aws_availability_zones.available`) instead of reading back module-created subnet IDs.
2. **Medium — ECS AZ alignment:** `gateway_ecs.private_subnet_ids` now uses `local.gateway_alb_subnet_ids` so tasks share the ALB’s filtered AZs when private origin is enabled.
3. **Medium — excluded AZ list:** Documented source/date (CloudFront VPC origins docs, 2026-07), noted `eu-west-1` has no exceptions, and noted the ValidationException failure mode if AWS adds exclusions.
4. **Low:** Removed the no-op data-source `moved` block; clarified the CF ingress-rule move is only for intermediate-branch stacks.
5. **Nit:** Trailing newline on `data.tf`. Left `origin_keepalive_timeout = 60` — matches every other origin in `cloudfront.tf`.

erikdw:
Follow-up from sandbox3 upgrade retest: CloudFront rejected **ApiEcsOrigin** creation because the API ECS ALB still spanned \`use1-az3\` (\`us-east-1e\` in this account). Gateway ALB filtering alone is not enough when \`enable_ecs_api\` creates a VPC origin.

Pushed a fix: shared \`cloudfront_vpc_origin_safe_subnet_ids\` used by API ECS ALB/tasks (and private gateway when enabled), plus an \`api_ecs_subnet_validation\` precondition.

erikdw:
Greenfield `erikdw-sandbox` plan failed with `Invalid count argument` on `aws_cloudfront_vpc_origin.gateway` because count depended on ALB ARN (unknown until apply). Fixed: count uses `use_private_ai_gateway_origin` only.

erikdw:
Minimized scope: **removed all CloudFront VPC-origin AZ filtering** (gateway + API ECS) from this PR.

Rationale: #287 already shipped \`ApiEcsOrigin\` without AZ handling; filtering both origins belongs in one follow-up. This PR is now just \`gateway-alb\` move + \`use_private_ai_gateway_origin\` / \`/v1/proxy\` wiring (+ greenfield \`count\` fix).

AGENTS.md
```diff
@@ -62,12 +63,14 @@ Module changes must be applyable directly to live customer stacks without tear-d
 
 ### Private gateway ALB relocation
 
-The gateway internal ALB moved from `gateway-ecs` to `services-common`. Stacks that already applied the old layout will destroy/recreate gateway ALB resources (new DNS). No `moved` blocks: private gateway is not in production use yet. Gateway ECS tasks may recycle when the target group is replaced.
+The gateway internal ALB lives in `modules/gateway-alb` so callers can reference `GATEWAY_URL` without depending on `gateway-ecs` (see cycle notes in PR / Kevin’s design). It previously lived in `gateway-ecs`, then briefly in `services-common`. `moved` blocks in `moved_state.tf` migrate state from `services_common` → `gateway_alb` so existing stacks (e.g. prod-eu) do not destroy/recreate the ALB. Those moved blocks can be removed after all customers have applied once.
```
KevinGreen: Nit: I think we should avoid specifics about existing deployments in these notes

modules/gateway-alb/versions.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 1.0"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 6.23.0, < 7.0.0"
+    }
+  }
+}
```
KevinGreen, line 9:
```suggestion
terraform {
  required_version = ">= 1.10.0"
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = ">= 6.23.0, < 7.0.0"
    }
  }
}
```
Just to be consistent with the other version.tf files

KevinGreen (approved):

PR #294, opened 2026-07-20
Add mocked plan-mode smoke tests for common deployment shapes

brianvans:
Inspired by https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/293, this adds some high level plan-time unit tests for common ways the module can be instantiated. Fixes a bug it caught in the process when `enable_brainstore=false`.

It's not a comprehensive level of coverage, but I'm hoping to just get something basic in place for us to verify if the value of having these around outweighs the overhead of maintaining the mocks.


main.tf
```diff
@@ -420,11 +420,11 @@ module "api_ecs" {
   function_tools_secret_arn = module.services_common.function_tools_secret_arn
 
   # Brainstore
-  brainstore_hostname             = module.brainstore[0].dns_name
-  brainstore_writer_hostname      = var.brainstore_writer_instance_count > 0 ? module.brainstore[0].writer_dns_name : null
-  brainstore_fast_reader_hostname = var.brainstore_fast_reader_instance_count > 0 ? module.brainstore[0].fast_reader_dns_name : null
-  brainstore_s3_bucket_name       = module.storage.brainstore_bucket_id
-  brainstore_port                 = module.brainstore[0].port
+  brainstore_hostname             = var.enable_brainstore ? module.brainstore[0].dns_name : null
```
KevinGreen: I think we can just get rid of enable_brainstore as a variable instead of fixing support for it. The data plane won't be able to run without it.

KevinGreen (dismissed):
This is great. Hopefully it'll help catch more errors like yesterday's

main.tf
brianvans: Opened https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/349 and cherry-picked the commit here so the tests pass

KevinGreen (approved):

PR #299, opened 2026-07-23
Fetch Brainstore service token from Secrets Manager

alexbasista-bt:
## Description 
Pass the function tools secret ARN to the Brainstore EC2 module instead of passing its plaintext value.

Brainstore reader/read-writer, writer, and fast-reader instances now retrieve `SERVICE_TOKEN_SECRET_KEY` from Secrets Manager during bootstrap (cloud-init), following the existing database-secret pattern. The Brainstore IAM role is granted `secretsmanager:GetSecretValue` access to the function tools secret.

Users do not need to provide any new inputs or change their configuration. Applying this change creates new Brainstore launch-template versions and triggers rolling instance refreshes for the corresponding Auto Scaling Groups.

## Context 
The service token was previously rendered directly into Brainstore EC2 launch-template user data. EC2 user data is base64-encoded rather than encrypted and can be retrieved through EC2 APIs by principals with the relevant permissions.

This change keeps the secret value out of launch-template user data by passing only its Secrets Manager ARN through Terraform. The instance retrieves the value at boot using its IAM role.

## Testing 
- Successfully deployed update via Terraform
- Confirmed cloud-init completed successfully and populated `SERVICE_TOKEN_SECRET_KEY` in the correct place
- Confirmed traces are being ingested after the change

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -565,7 +565,7 @@ module "brainstore" {
   use_redis_replication_group           = var.use_redis_replication_group
   redis_host                            = module.redis.redis_endpoint
   redis_port                            = module.redis.redis_port
-  service_token_secret_key              = module.services_common.function_tools_secret_key
+  service_token_secret_arn              = module.services_common.function_tools_secret_arn
   brainstore_s3_bucket_arn              = module.storage.brainstore_bucket_arn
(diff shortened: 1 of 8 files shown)


erikdw (approved):
🚢 🇮🇹

PR #300, opened 2026-07-23
Add Redis auth token support and securely inject Redis URLs

thejohnny:
## Description 
Extends security model for Elasticache Redis replication groups to require a Redis auth token to connect. For ECS-based services, inject fully rendered `REDIS_URL` from Secrets Manager. Render authenticated Redis URL for Lambda services.

## Context 
Unblock Enterprise customer requiring authenticated Redis endpoints.

## Testing 
Deployed new Lambda- and ECS-based data planes using Elasticache replication group. Modified Lamba-based data plane to ECS. Ran `braintest run evaltest` and created/ran playground in each deployment


modules/elasticache/main.tf
```diff
@@ -58,6 +59,42 @@ resource "aws_elasticache_replication_group" "main" {
   transit_encryption_enabled = true
   transit_encryption_mode    = "required"
 
+  auth_token                 = local.redis_auth_token
+  auth_token_update_strategy = "SET"
```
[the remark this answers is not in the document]
thejohnny: Changed to ROTATE

modules/brainstore-ec2/templates/user_data.sh.tpl
```diff
@@ -127,6 +127,8 @@ DB_CREDS=$(aws secretsmanager get-secret-value --secret-id ${database_secret_arn
 DB_USERNAME=$(echo $DB_CREDS | jq -r .username)
 DB_PASSWORD=$(echo $DB_CREDS | jq -r .password)
 
+# Get the fully-rendered Redis URL from Secrets Manager
+REDIS_URL=$(aws secretsmanager get-secret-value --secret-id ${redis_url_secret_arn} --query SecretString --output text)
```
ibolmo: should we try the env vars first and if unset then we try secrets manager?

modules/gateway-ecs/main.tf
```diff
@@ -64,6 +61,16 @@ locals {
         value = local.merged_env_vars[key]
       }
     ]
+    secrets = [
+      {
+        name      = "COMPLETIONS_CACHE_REDIS_URL"
+        valueFrom = var.redis_url_secret_arn
```
ibolmo, line 67: #nit? should gateway be using the same redis?

modules/services/main.tf
```diff
@@ -53,7 +53,16 @@ locals {
     lambda => trimspace(data.http.lambda_versions[lambda].response_body)
   }
 
-  postgres_url                 = "postgres://${var.postgres_username}:${var.postgres_password}@${var.postgres_host}:${var.postgres_port}/postgres"
+  postgres_url = "postgres://${var.postgres_username}:${var.postgres_password}@${var.postgres_host}:${var.postgres_port}/postgres"
```
ibolmo, line 56: curious why we couldn't use the secrets manager here like in other parts?

modules/brainstore-ec2/variables.tf
```diff
@@ -85,19 +85,9 @@ variable "database_port" {
...
-  description = "Sets BRAINSTORE_REDIS_URI and BRAINSTORE_XACT_MANAGER_URI with the appropriate scheme (rediss:// for replication group, redis:// for legacy cluster)"
-}
-
-variable "redis_host" {
-  type        = string
-  description = "The hostname of the Redis instance"
-}
-
-variable "redis_port" {
+variable "redis_url_secret_arn" {
```
ibolmo, line 88: same question whether this should be the same redis

ibolmo (dismissed):

modules/elasticache/main.tf
```diff
@@ -58,6 +59,42 @@ resource "aws_elasticache_replication_group" "main" {
   transit_encryption_enabled = true
   transit_encryption_mode    = "required"
 
+  auth_token                 = local.redis_auth_token
+  auth_token_update_strategy = "ROTATE"
```
erikdw: If Cursor's review is to be trusted, we'll want a 2-phased strategy to move through ROTATE to SET:

> [!Warning]
> **Permanent `ROTATE` does not require AUTH on existing stacks**
> 
> For upgrades of live unauthenticated replication groups, AWS leaves passwordless access open after `ROTATE` until a follow-up `SET`. New creates are fine. Need a documented/automated two-phase path (`ROTATE` → roll clients → `SET`), or a variable to flip strategy.

modules/brainstore-ec2/templates/user_data.sh.tpl
erikdw: Would prefer Secrets Manager only here (no env fallback) and fail closed like `SERVICE_TOKEN_SECRET_KEY` below.

In general we wanna keep secret _values_ out of user_data. So the pattern should be only ARNs in the template, resolve at boot. 

There are existing secrets we need to get out of user_data (Zander just excised one last week), just don't wanna extend it to be more work.

erikdw (changes requested):

modules/brainstore-ec2/templates/user_data.sh.tpl
thejohnny: Script will now fail if `REDIS_URL` cannot be fetched.

modules/gateway-ecs/main.tf
thejohnny: Good callout. If enabled, the current [gateway_ecs](https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/blob/e7f50a312390d47a4587bae86767a4ac8e72ec0b/main.tf#L364-L365) module invocation shares the same Redis as the rest of the deployment. This change maintains that architecture but changes how the secret is managed.

modules/services/main.tf
thejohnny: we can and should use the secrets ARN pattern. will update separately.

modules/elasticache/main.tf
thejohnny: Added a variable to provide a path from `ROTATE` to `SET`

erikdw (dismissed):
This LGTM! :shipit: 

Nice work on the Secrets Manager path and the `ROTATE`/`SET` variable.

One minor ask: please document the existing-cluster upgrade: apply `ROTATE` → cut clients over → apply `SET`. Stopping after `ROTATE` leaves passwordless open while the stack looks healthy, which is easy to miss without a runbook.  Maybe that should be done by our doc-team.

erikdw:
Thanks for bearing with me on this review @thejohnny. Looks like the at-rest encryption change and the existing auth-token ignore_changes each need a small README follow-up.

### 1. Existing RG upgrades replace Redis
`at_rest_encryption_enabled` and `kms_key_id` are both `ForceNew` in the AWS provider. On `main`, replication groups only had transit encryption, so adding at-rest + the module KMS key will replace any existing `use_redis_replication_group = true` stack on upgrade. So this is a Redis replacement, not an in-place auth enable.

The README already warns about legacy cluster → RG downtime. Please also call out that **already-on-RG customers hit a Redis replacement** when picking up this module version, for the same reason.

### 2. README rotation path is incomplete
Docs say: set `ROTATE`, deploy, rotate token, set `SET`. But `aws_secretsmanager_secret_version.auth_token` has `ignore_changes = [secret_string]`, so Terraform won’t apply a new token on its own. 

Either document the manual secret/ElastiCache steps, or say rotation isn’t supported via the module yet.

PR #302, opened 2026-07-24
Opt-in quarantine → private gateway via PrivateLink (NLB→ALB)

erikdw:
## Summary

This change adds `use_private_gateway_quarantine_proxy`. The default is `false`.

When the flag is `false`, Terraform does not create PrivateLink. Terraform does not auto-wire quarantine to the private gateway.

When `quarantine_proxy_url` is set, that URL always wins.

When the flag is `false` and there is no override, quarantine uses the in-account AI Proxy Lambda Function URL.

When the flag is `true`, Terraform can create a PrivateLink path from quarantine to the private gateway. That path uses an NLB, then the gateway ALB, then an interface VPC endpoint in quarantine. This path does not use VPC peering.

Quarantine reaches the VPC endpoint. Quarantine does not talk to the gateway ALB directly. The gateway ALB security group allows the PrivateLink NLB security group on port 80.

Terraform creates PrivateLink only when all of these are true:

- `use_private_gateway_quarantine_proxy` is `true`.
- `create_ai_gateway` is `true`.
- The module creates the main VPC (`create_vpc`).
- The module creates the quarantine VPC (`enable_quarantine_vpc` and no `existing_quarantine_vpc_id`).
- `use_global_ai_gateway_origin` is `false`.

When PrivateLink exists and there is no override, `QUARANTINE_PROXY_URL` is `http://<vpce-dns>/v1/proxy`.

Loop uses `self_hosted_ai_proxy_url`. That value is the AI Proxy Lambda Function URL. PrivateLink does not change Loop.

### URL precedence

Terraform selects `QUARANTINE_PROXY_URL` in this order:

1. If `quarantine_proxy_url` is set, Terraform uses that URL. This override always wins.
2. If Terraform created the PrivateLink VPC endpoint, Terraform uses `http://<vpce-dns>/v1/proxy`.
3. In all other cases, Terraform uses the in-account AI Proxy Lambda Function URL.

### PrivateLink pieces

When Terraform creates PrivateLink, it creates these pieces:

- The provider NLB is an internal multi-AZ NLB in the main private subnets. Its security group allows port 80 for PrivateLink consumers.
- The target group uses `target_type = "alb"` and points to the gateway ALB. The listener is TCP on port 80. The health check is `HTTP /health`.
- The endpoint service is `aws_vpc_endpoint_service` with `acceptance_required = false` for the same account.
- The consumer VPC endpoint is an interface endpoint in the quarantine private subnets. Its security group allows the quarantine Lambda security group on port 80.
- The gateway ALB security group allows ingress from the NLB security group on port 80. It does not allow the quarantine CIDR or a cross-VPC Lambda security group for this path.
- The URL is `http://<vpce-dns>/v1/proxy`. Traffic uses HTTP on this hop. Traffic stays inside AWS PrivateLink.

### Cost and AZ notes

PrivateLink adds an internal NLB. The NLB has hourly cost and LCU cost. The interface endpoint also has a cost.

If the NLB and the VPC endpoint ENIs do not cover the same AZs as the gateway ALB, the endpoint can stay unhealthy or unresolved.

### Existing VPC or existing quarantine

Terraform does not create the consumer VPC endpoint when you use `existing_vpc_id` or `existing_quarantine_vpc_id`.

1. Create an interface endpoint to `quarantine_gateway_privatelink_service_name`.
2. Set `quarantine_proxy_url` to `http://<your-vpce-dns>/v1/proxy`.

This change does not depend on braintrust [#18013](https://github.com/braintrustdata/braintrust/pull/18013).

### Outputs

The module exposes these values:

- `quarantine_proxy_url`: The URL that API ECS uses for `QUARANTINE_PROXY_URL`.
- `quarantine_gateway_privatelink_service_name`: The endpoint service name. This is null unless Terraform created PrivateLink.
- `quarantine_gateway_privatelink_endpoint_dns_name`: The DNS name of the quarantine VPC endpoint. This is null unless Terraform created the consumer.

## Test plan

- [x] If `use_private_gateway_quarantine_proxy` is `false` and `quarantine_proxy_url` is unset, the URL is the Function URL. Terraform does not create the NLB, the endpoint service, or the VPC endpoint.
- [x] If `quarantine_proxy_url` is set, the override wins. If the flag is `false`, Terraform still does not create PrivateLink.
- [x] If the flag is `true` with module VPCs, `create_ai_gateway`, and no global origin, the URL is `http://<vpce-dns>/v1/proxy`. The NLB, the endpoint service, and the VPC endpoint exist. Terraform does not create peering or extra routes. The ALB allows only the NLB security group for this path.
- [x] If `use_global_ai_gateway_origin` is `true`, the URL is the Function URL. Terraform does not create PrivateLink.
- [x] If `enable_ecs_api` is `false`, the fallback URL is still the AI Proxy Function URL.
- [x] If the plan uses an existing VPC or an existing quarantine VPC and the flag is `true`, Terraform does not create the consumer endpoint. Operators must create a manual endpoint. Operators must set `quarantine_proxy_url`.
- [x] Sandbox apply with the flag and quarantine on. UDF LLM calls use the private gateway through the VPC endpoint. Calls do not use `gateway.braintrust.dev`. Calls do not use the gateway ALB DNS.
- [x] `terraform validate` passed on the module and the production example.


erikdw:
Addressed Claude review feedback on this PR:

**Fixed**
- Restored `main_vpc_private_route_table_id` (matches `main`; new PrivateLink outputs kept alongside it)
- Removed `braintrust_api_container_image_repository` root var + `main.tf` coalesce (sandbox private ECR stays local-only / untracked under `*.deployed/`)
- Omit `QUARANTINE_PROXY_URL` from the ECS task env when null/blank (instead of baking `""`); description updated to match. Aligns with api-ts `getRuntimeEnv` fallback behavior
- Dropped NLB SG `0.0.0.0/0 :80` ingress and documented why `enforce_security_group_inbound_rules_on_private_link_traffic = "off"` admits PrivateLink traffic
- Noted intentional plain HTTP on the PrivateLink hop in `AGENTS.md`

**Deferred**
- Codex P2 about gating quarantine URL validation on `enable_quarantine_vpc` (separate from Claude’s list; can follow up if needed)

`mise run validate` succeeded (module + production example).

erikdw:
## Codex local review fixes

Addressed the two Codex P1/P2 findings on this branch:

**P1 — endpoint service allowed principal**
- Added `aws_vpc_endpoint_service_allowed_principal` for the current account root (`arn:aws:iam::<account_id>:root` via `data.aws_caller_identity.current`)
- Consumer `aws_vpc_endpoint.quarantine_gateway` now `depends_on` that allow-list so create order is correct
- `acceptance_required = false` alone is not enough; AWS still requires discoverability via allowed principals

**P2 — APN tags**
- Switched quarantine PrivateLink tag merges from `var.custom_tags` → `local.all_custom_tags` (NLB, SGs, endpoint service, endpoint, SG rules, target group) so the forced APN partner tag is applied

`terraform fmt` + `mise run validate` both succeeded.

erikdw:
Addressed Codex P2: `aws_lb_target_group_attachment.gateway_quarantine_privatelink_alb` now `depends_on = [module.gateway_alb]` so NLB alb-target registration waits until the gateway ALB (and its HTTP listener) exist, avoiding a greenfield RegisterTargets race.

KevinGreen (approved):

PR #307, opened 2026-07-30
[INF-178] Fetch Brainstore license and Datadog keys from Secrets Manager

erikdw:
## Summary
- Store Brainstore license key and internal observability API key in Secrets Manager instead of embedding them in Brainstore EC2 user_data / gateway ECS plain env vars.
- Brainstore boots fetch both secrets (version IDs baked into user_data so rotations still force rolling refresh); gateway injects `BRAINSTORE_LICENSE_KEY` via ECS task secrets.
- Mark root `brainstore_license_key` and `internal_observability_api_key` as sensitive.
- CI also validates the external-EKS example with `brainstore_license_key` unset (wiring/syntax coverage only — `validate` does not catch eval-time coalesce-on-null).

https://linear.app/braintrustdata/issue/INF-178

**Follow-up:** plan-mode `terraform test` for the null license path → #308

## Test plan
- [ ] `mise run lint` / `mise run validate` (includes external-EKS example with license key unset — wiring only)
- [ ] Plan against a sandbox with Brainstore (+ optional private gateway): confirm new license secret create, user_data no longer contains plaintext keys, gateway task def uses `secrets` for license
- [ ] Apply and verify Brainstore instances boot with license + (if set) Datadog agent; gateway has `BRAINSTORE_LICENSE_KEY` from SM
- [ ] Rotate license / observability key via TF var and confirm secret version bump triggers Brainstore rolling refresh / ECS rollout


erikdw:
**Follow-up:** plan-mode regression coverage for unset `brainstore_license_key` is staged separately — `validate` alone cannot catch the coalesce class of bug.

→ https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/308

PR #311, opened 2026-07-31
Add Loop runtime module (ECS service + MicroVM sandbox)

alexr17:
## Context

Adds support for deploying the Loop runtime — the service behind hosted Loop — to the data plane. It is gated behind a new `enable_loop_runtime` flag that defaults to off, so existing deployments are unaffected until they opt in.

## Description

Three new modules that follow the same patterns as the existing service modules:

- `loop-runtime-alb` — internal load balancer in front of the service.
- `loop-runtime-ecs` — the Fargate service, task definition, autoscaling, and IAM.
- `loop-runtime-sandbox-aws-microvm` — builds the AWS Lambda MicroVM image the service uses to run sandboxed code. It is wired to the service module through a small interface so other sandbox backends can be added later without touching the service module.

Root variables add the `enable_loop_runtime` flag and related `loop_runtime_*` settings, and `/loop/runtime*` is routed through the existing CloudFront distribution. The MicroVM image is created through a small nested CloudFormation stack because the AWS Terraform provider does not yet have a native resource for it.

The MicroVM sandbox defaults to AWS-managed Internet egress. Setting `loop_runtime_sandbox_egress_mode` to any other value provisions a dedicated restricted path with no outbound network access.

## Testing

`terraform validate` and `terraform fmt` pass across the root module and every new submodule. The MicroVM image build was exercised end to end (build succeeds, image reaches the ready state) and torn down.

### DIFF
--- a/loop-runtime.tf
+++ b/loop-runtime.tf
@@ -1,0 +1,130 @@
+locals {
+  # Loop runtime requires the ECS API data plane and Brainstore (it reads
+  # module.brainstore[0] for the reader URL / SG). It is fronted by the shared
+  # CloudFront distribution at /loop/runtime*.
+  create_loop_runtime = local.create_ecs_api && var.enable_brainstore && var.enable_loop_runtime
+
+  loop_runtime_version = (
+    var.loop_runtime_version_override != null
+    ? var.loop_runtime_version_override
+    : jsondecode(file("${path.module}/modules/loop-runtime-ecs/VERSIONS.json"))["loop-runtime"]
+  )
+
+  loop_runtime_brainstore_reader_url = local.create_loop_runtime ? format(
+    "http://%s:%s",
+    var.brainstore_fast_reader_instance_count > 0 ? module.brainstore[0].fast_reader_dns_name : module.brainstore[0].dns_name,
+    module.brainstore[0].port,
+  ) : ""
+}
+
+module "loop_runtime_alb" {
+  source = "./modules/loop-runtime-alb"
+  count  = local.create_loop_runtime ? 1 : 0
+
+  deployment_name                      = var.deployment_name
+  vpc_id                               = local.main_vpc_id
+  private_subnet_ids                   = local.main_vpc_private_subnet_ids
+  enable_cloudfront_vpc_origin_ingress = true
+  authorized_security_groups = {
+    "API"        = module.services_common.api_security_group_id
+    "Brainstore" = module.services_common.brainstore_instance_security_group_id
+  }
+  alb_deregistration_delay = var.loop_runtime_alb_deregistration_delay
+  custom_tags              = local.all_custom_tags
+}
+
+module "loop_runtime_sandbox_aws_microvm" {
+  source = "./modules/loop-runtime-sandbox-aws-microvm"
+  count  = local.create_loop_runtime ? 1 : 0
+
+  deployment_name          = var.deployment_name
+  permissions_boundary_arn = var.permissions_boundary_arn
+  microvm_version_tag      = local.loop_runtime_version
+
+  microvm_minimum_memory_mib            = var.loop_runtime_microvm_minimum_memory_mib
+  microvm_max_idle_duration_seconds     = var.loop_runtime_microvm_max_idle_duration_seconds
+  microvm_suspended_duration_seconds    = var.loop_runtime_microvm_suspended_duration_seconds
+  microvm_maximum_duration_seconds      = var.loop_runtime_microvm_maximum_duration_seconds
+  microvm_auth_token_expiration_minutes = var.loop_runtime_microvm_auth_token_expiration_minutes
+  enable_microvm_runtime_logs           = var.enable_loop_runtime_microvm_runtime_logs
+
+  kms_key_arn = local.kms_key_arn
+  custom_tags = local.all_custom_tags
+}
+
+module "loop_runtime_ecs" {
+  source = "./modules/loop-runtime-ecs"
+  count  = local.create_loop_runtime ? 1 : 0
+
+  deployment_name    = var.deployment_name
+  kms_key_arn        = local.kms_key_arn
+  vpc_id             = local.main_vpc_id
+  private_subnet_ids = local.main_vpc_private_subnet_ids
+  ecs_cluster_arn    = module.ecs[0].cluster_arn
+  ecs_cluster_name   = module.ecs[0].cluster_name
+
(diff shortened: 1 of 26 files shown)


modules/loop-runtime-ecs/main.tf
```diff
@@ -0,0 +1,622 @@
...
+  name = "LoopRuntimeBrainstoreS3Access"
+  role = aws_iam_role.task.id
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = concat(
+      [
+        {
+          Effect   = "Allow"
+          Action   = ["s3:PutObject"]
+          Resource = ["${var.brainstore_s3_bucket_arn}/brainstore/wal/object-store-objects/*"]
```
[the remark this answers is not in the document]
alexr17: this is not necessary

modules/loop-runtime-ecs/VERSIONS.json
```diff
@@ -0,0 +1,3 @@
+{
+    "loop-runtime": "latest"
```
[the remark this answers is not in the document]
alexr17: i will add this once the next release comes out next week

loop-runtime.tf
```diff
@@ -0,0 +1,130 @@
...
+  source = "./modules/loop-runtime-alb"
+  count  = local.create_loop_runtime ? 1 : 0
+
+  deployment_name                      = var.deployment_name
+  vpc_id                               = local.main_vpc_id
+  private_subnet_ids                   = local.main_vpc_private_subnet_ids
+  enable_cloudfront_vpc_origin_ingress = true
+  authorized_security_groups = {
+    "API"        = module.services_common.api_security_group_id
+    "Brainstore" = module.services_common.brainstore_instance_security_group_id
```
brianvans: Does Brainstore need access or is just api sufficient?

modules/loop-runtime-ecs/main.tf
```diff
@@ -0,0 +1,622 @@
...
+}
+
+# --- Task execution role (pulls secrets at container start) ---
+resource "aws_iam_role" "task_execution" {
+  name               = "${var.deployment_name}-loop-runtime-task-exec"
+  assume_role_policy = data.aws_iam_policy_document.ecs_task_assume_role.json
+  tags = merge({
+    Name = "${var.deployment_name}-loop-runtime-task-exec"
+  }, local.common_tags)
+}
```
brianvans, line 337: Does this role need a permissions_boundary? We claim all roles will have one [here](https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/blob/ac70cfe3f822600678a01c08f11229bc3444c8e6/variables.tf#L1286)

modules/loop-runtime-alb/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+  validation {
+    condition     = var.alb_idle_timeout >= 1 && var.alb_idle_timeout <= 4000
+    error_message = "alb_idle_timeout must be between 1 and 4000 seconds."
+  }
+}
+
+variable "alb_deregistration_delay" {
+  type        = number
+  description = "Seconds for the Loop runtime target group to wait before deregistering draining targets."
+  default     = 900
```
brianvans, line 58: I'm assuming this extended drain period is intentional to support streaming connections, but just want to double check, as it will make terraform applies take quite a while.

brianvans (commented):
Applied cleanly to sandbox 👍. Left a few comments, mostly around loop-specific things I don't understand as well as I probably should. Wouldn't mind getting another set of 👀 and terraform plan against a few other environments before final ✅ 

Also, should we add any of the loop runtime's outputs to top-level outputs.tf?

modules/loop-runtime-ecs/main.tf
alexr17: thanks great catch

loop-runtime.tf
alexr17: good catch, it doesn't need to be this permissive

modules/loop-runtime-alb/variables.tf
alexr17: yeah it is. 900s may feel long but it's just extra cautious

modules/loop-runtime-sandbox-aws-microvm/main.tf
```diff
@@ -0,0 +1,269 @@
...
+    ]
+  })
+}
+
+# MicroVM image built via a scoped nested CloudFormation stack. The AWS provider
+# has no native AWS::Lambda::MicrovmImage resource, and the awscc provider cannot
+# create it (Cloud Control prunes the empty-but-required EgressNetworkConnectors/
+# EnvironmentVariables/Hooks keys -> 400). CloudFormation accepts them, matching
+# the BYOC template semantics.
+resource "aws_cloudformation_stack" "microvm_image" {
```
KevinGreen, line 406: This feels a little weird. Is it possible for us to build and release the microvm image once instead of having to build it for each deployment?

alexr17: yeah the first problem is that terraform does not support microvms yet. that's why there's this hacky cloudformation thing.

the second part is that we're hoping for a simple deployment process but the current microvm image deployment behavior is not robust enough to do rollbacks well. i'll dig more into this and see if we can avoid doing it this way.

jeffmccollum: The awscc provider has support for it
https://github.com/hashicorp/terraform-provider-awscc/releases/tag/v1.90.0
Example 
https://github.com/mkdev-me/terraform-aws-github-runner-lambda-microvms/blob/main/image.tf

Looks like there's a PR in progress for the regular provider 
https://github.com/hashicorp/terraform-provider-aws/pull/48984

alexr17: yeah terraform support is gonig slowly

but awscc doesn't support these options which causes 400s. so if awscc adds these we can leverage it.
  - EgressNetworkConnectors: []
  - EnvironmentVariables: []
  - Hooks: { MicrovmHooks: {}, MicrovmImageHooks: {} }

alexr17: okay so i explored this for quite a while and this is cleanest way to keep parity with our existing CF logic. i think it's probably best to just wait for native terraform support and handle this once that lands.

i will file a follow up for this

KevinGreen (approved):

PR #318, opened 2026-08-06
[INF-196] Add opt-in allowlist for Bedrock Gateway AssumeRole

erikdw:
## Summary
- Adds `ai_gateway_bedrock_assume_role_arns` so self-hosted customers can opt in to bounding the AI Gateway task role `sts:AssumeRole` used for Bedrock AssumeRole auth
- Empty default keeps `Resource: "*"`; when set, only listed role ARNs / IAM Resource patterns may be assumed (`ExternalId: bt:*` unchanged)
- Stacked on #317 (export allowlist); does not change hosted `infra` Gateway

## Test plan
- [x] `mise run lint`
- [x] Confirm empty `ai_gateway_bedrock_assume_role_arns` leaves Gateway AssumeRole unrestricted
- [x] Confirm a non-empty allowlist appears on the gateway task role policy when `create_ai_gateway` is true
- [x] Confirm no change when `create_ai_gateway` is false

Verified against `erikdw-sandbox` (`examples/braintrust-data-plane-sandbox-private-gateway.deployed`, plan-only):
- Empty allowlist: `gateway_ecs` `customer_assume_role` is no-op with `Resource: "*"`
- Non-empty allowlist: planned update sets `Resource` to the exact ARN + pattern list
- `create_ai_gateway=false`: `customer_assume_role` is deleted with the gateway module (allowlist does not update an existing policy in place)

Stacked on https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/317
Part of https://linear.app/braintrustdata/issue/INF-196

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -1193,6 +1193,27 @@ variable "s3_export_assume_role_arns" {
   }
 }
 
+variable "bedrock_assume_role_arns" {
+  type        = list(string)
+  description = <<-EOT
+    Optional allowlist of IAM role ARNs that the AI Gateway task role may assume for Bedrock AssumeRole auth (sts:AssumeRole with ExternalId bt:*).
+    When empty (default), AssumeRole remains unrestricted (Resource "*") for backward compatibility.
+    When set, only matching role ARNs may be assumed. Exact ARNs and IAM Resource patterns are both accepted, for example:
+      ["arn:aws:iam::123456789012:role/braintrust-bedrock-role"]
+    Decision (reversible): wildcards are allowed so customers can use naming-convention patterns; we can tighten to exact ARNs later if needed.
+    Only applies when create_ai_gateway is true.
+  EOT
+  default     = []
+
+  validation {
+    condition = alltrue([
+      for arn in var.bedrock_assume_role_arns :
+      can(regex("^arn:aws:iam::(\\*|[0-9]{12}):role/.+$", arn))
+    ])
+    error_message = "bedrock_assume_role_arns entries must be IAM role ARNs or ARN patterns of the form arn:aws:iam::<account-id|*>:role/<role-name-or-pattern>."
+  }
+}
+
 variable "brainstore_s3_bucket_retention_days" {
   type        = number
   description = "The number of days to retain non-current S3 objects. e.g. deleted objects"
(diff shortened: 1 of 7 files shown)


variables.tf
```diff
@@ -1194,6 +1194,27 @@ variable "s3_export_assume_role_arns" {
...
 }
 
+variable "bedrock_assume_role_arns" {
+  type        = list(string)
+  description = <<-EOT
+    Optional allowlist of IAM role ARNs that the AI Gateway task role may assume for Bedrock AssumeRole auth (sts:AssumeRole with ExternalId bt:*).
+    When empty (default), AssumeRole remains unrestricted (Resource "*") for backward compatibility.
+    When set, only matching role ARNs may be assumed. Exact ARNs and IAM Resource patterns are both accepted, for example:
+      ["arn:aws:iam::123456789012:role/braintrust-bedrock-role"]
+    Decision (reversible): wildcards are allowed so customers can use naming-convention patterns; we can tighten to exact ARNs later if needed.
```
brianvans: Not sure I follow this "decision (reversible)" part. Is this a duplicate of the wildcard stuff from above or different somehow?

variables.tf
```diff
@@ -1194,6 +1194,27 @@ variable "s3_export_assume_role_arns" {
   }
 }
 
+variable "bedrock_assume_role_arns" {
```
brianvans: If this is only for the gateway, we might consider prefixing it with `ai_gateway_` to make that immediately obviously like we do for some of the other gateway-specific variables

brianvans (dismissed):
Looks good overall! Left two comments but nothing major

variables.tf
erikdw: ah, yeah, this is the clanker's way of saying that we **_could_** choose to have strict enforcement of explicit ARNs and resource patterns.  But that feels too strict.

Should probably remove that line from the docs, what do ya think?  

Same applies in this sibling PR: https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/317

variables.tf
erikdw: Good catch, consistency is king.

KevinGreen (approved):

PR #321, opened 2026-08-07
Default ECS quarantine proxy to CloudFront hairpin

erikdw:
## Summary
- Split API ECS ALB into `modules/api-ecs-alb` (with `moved` blocks) so tasks can bake `ingress.api_url` into `QUARANTINE_PROXY_URL` without a Terraform cycle.
- When `enable_ecs_api` is on, quarantine LLM calls default to the dataplane public `/v1/proxy` CloudFront hairpin (`custom_domain` preferred, else `*.cloudfront.net`), instead of hosted `gateway.braintrust.dev`.
- Add optional `quarantine_proxy_url` override for hosted gateway / custom URLs.

## Test plan
- [ ] **Untested so far** — needs a sandbox apply before merge
- [ ] `terraform plan` on an existing ECS dataplane: ALB resources show as moves (not destroy/create)
- [ ] Confirm `QUARANTINE_PROXY_URL` on API ECS is `https://<custom_domain|cloudfront>/v1/proxy`
- [ ] Invoke a quarantine UDF that calls an LLM; traffic reaches dataplane `/v1/proxy` (not hosted gateway)
- [ ] Override with `quarantine_proxy_url = "https://gateway.braintrust.dev/v1/proxy"` and confirm env flips
- [ ] Rollback path: unset override / revert module bump still healthy


erikdw:
This is defunct. We use the PrivateLink NLB bridge from #302 instead.

PR #327, opened 2026-08-12
Use stable METHOD:path keys for API ECS ALB listener rules

brianvans:
## Description

Refactors the internal representation of listener rules. Adds a new routing rule for POST /automation/trigger in the process.

## Context

This change updates Terraform resource identity to support more stable listener-rule management for future module changes.

In the forward direction, this change is a no-op. ~everting to an earlier module version after applying this change will cause listener rules to be destroyed and recreated. For deployments using `enable_ecs_api=true`, this would mean a brief period of time where all traffic routes to the main API pool rather than.

## Testing

- Verified that the forward upgrade is a no-op for existing listener rules.
- Verified that the previous module version may plan listener-rule recreation after upgrading to this version.
- Confirmed that listener-rule behavior and routing configuration are unchanged during the forward upgrade.


modules/api-ecs/alb-path-routes.tf
```diff
@@ -31,7 +34,7 @@ locals {
 
   alb_path_listener_rules = {
     for idx, route in local.alb_path_routes :
-    tostring(idx + 1) => merge(
+    "${try(route.method, "ANY")}:${route.path}" => merge(
       { method = null },
       route,
       { priority = idx + 1 },
```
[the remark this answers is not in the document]
brianvans: Added a key for priority to the objects in `alb_path_routes` so we are explicit about priorities rather than relying on implicit ordering of the list

KevinGreen (dismissed):

modules/api-ecs/alb-path-routes.tf
```diff
@@ -1,40 +1,50 @@
...
+# for_each keys are stable METHOD:path strings so adding or reordering routes
+# does not reshuffle keys and force listener-rule replace.
+#
+# Priorities are explicit (not idx+1). The AWS provider updates
+# aws_lb_listener_rule.priority via SetRulePriorities for that one rule only;
+# unspecified rules keep their current priority. Index-derived priorities would
+# shift neighbors on insert and fail with PriorityInUse. Keep the historical
+# 1..N values here so this upgrade only renames for_each keys (via moved
+# blocks) without priority churn. When adding a route, assign an unused
+# priority (append with max+1, or renumber with gaps if inserting mid-list).
```
erikdw: The clankers think it's *slightly* more nuanced if we need to renumber.  They suggest this comment change:

```suggestion
# blocks) without priority churn.
# When adding a route, assign an unused priority (max+1). Do not renumber
# existing rules in the same apply because SetRulePriorities is per-rule and
# will fail with PriorityInUse. To change evaluation order, use a two-phase
# apply (move to a free high range, then to the final values).
```

erikdw (changes requested):

modules/api-ecs/alb-path-routes.tf
brianvans: Great point. I'm not sure that a two-phase change is practical for most purposes, short of a major module release and a big warning in the migration docs.

I wonder if now is a good time to re-number the priorities like we are programming Fortran (110,120,130) and leave a warning to avoid re-numbering existing priorities unless absolutely necessary.

erikdw: The N-phase apply requirement is definitely annoying. But I'll note that we already have such requirements for various features being enabled/rolled out for existing dataplanes when they are upgraded (e.g., the ECS enablement, and Gateway enablement).  So we do already document situations where people (and thus our automations) would need to do multiple phases.  Though I think we wanna keep the "greenfield first deployment" as a 1-pass apply unless we absolutely cannot.

All that being said, I like the idea of leaving some gaps between the priorities.

brianvans: Documented that specific gotcha and switched the priorities to have gaps for future rules.

We might want to target a major release with this change, since the revert path will be a delete+re-create of the rules causing all traffic to hit the default pool for a brief window.

erikdw: that's definitely a pretty big footgun yeah. Doing a major release sounds like a nicely operator-sympathetic move, but maybe we can bundle with some other big changes (e.g., making ECS mode the default).

erikdw (dismissed):

soldatchenko (approved):

PR #328, opened 2026-08-12
Opt-in s3 access logs

brianvans:
## Description

Opt-in S3 server access logging for the brainstore, code-bundle, and lambda-responses buckets. Disabled by default. When enabled, logs are written to a customer-owned bucket in the same account and region.

Operators can enable it with:

```hcl
s3_server_access_logging_bucket = "your-audit-logs-bucket"
# optional; defaults to "<deployment_name>/"
# s3_server_access_logging_prefix = "braintrust/"
```

The destination bucket must grant `s3:PutObject` to `logging.s3.amazonaws.com`, use SSE-S3 (not SSE-KMS), and must not have Object Lock. See the README for the required bucket policy. New outputs `code_bundle_s3_bucket_name` and `lambda_responses_s3_bucket_name` are available for that policy.

Existing deployments are unchanged unless the new variable is set.

> [!WARNING]
> Brainstore in a busy environment can generate a significant amount of s3 access logs

## Context

A Vanilla Hybrid customer needs server access logs from these buckets delivered to a bucket they own, for audit/compliance. This keeps logging optional so other customers are not affected.

## Testing

- [x] `terraform fmt` and `terraform validate` on the root module
- [x] `tflint` on the root and storage modules
- [x] Exercise an example configuration in sandbox environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,9 +119,11 @@ See the [`examples/cloudfront-logging`](examples/cloudfront-logging) directory f
 S3 server access logging is disabled by default. Enable it to deliver access logs from the brainstore, code-bundle, and lambda-responses buckets to an S3 bucket you own. This is commonly used for audit and compliance requirements.
 
 ```hcl
-s3_server_access_logging_bucket = "your-audit-logs-bucket"
-# Optional. Defaults to "<deployment_name>/". Per-bucket suffixes are appended.
-# s3_server_access_logging_prefix = "braintrust/"
+s3_server_access_logging = {
+  bucket = "your-audit-logs-bucket"
+  # Optional. Defaults to "<deployment_name>/". Per-bucket suffixes are appended.
+  # prefix = "braintrust/"
+}
 ```
 
 The destination bucket must:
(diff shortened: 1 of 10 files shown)


modules/storage/s3-logging.tf
```diff
@@ -0,0 +1,13 @@
+resource "aws_s3_bucket_logging" "access_logs" {
+  for_each = local.s3_server_access_logging_enabled ? local.s3_server_access_logging_buckets : {}
+
+  bucket        = each.value
+  target_bucket = var.s3_server_access_logging.bucket
```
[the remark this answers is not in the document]
brianvans: The destination bucket and policy are customer-owned and outside the module graph, so it's not really practical to do this. I updated the docs to make the expected order of operations more apparent

jeffmccollum (dismissed):

KevinGreen (dismissed):

README.md
```diff
@@ -114,6 +114,73 @@ If you need to enable CloudFront standard access logging, you can configure it i
...
+1. Create the destination bucket (same AWS account and region as the data plane). It must not have Object Lock or Requester Pays enabled, and default encryption must be SSE-S3 (AES256). SSE-KMS prevents Amazon S3 from delivering logs you can decrypt.
+2. Deploy the data plane (or use an existing deployment) so the source bucket name outputs are available.
+3. Attach a bucket policy on the destination bucket that grants `s3:PutObject` to `logging.s3.amazonaws.com` (see below). Restrict access to the destination bucket; access logs can include object keys and requester information.
+4. Only after that policy is applied, set `s3_server_access_logging` and apply again:
+
+```hcl
+s3_server_access_logging = {
+  bucket = "your-audit-logs-bucket"
+  # Optional. Defaults to "<deployment_name>/". Per-bucket suffixes are appended.
+  # prefix = "braintrust/"
```
erikdw: The destination policy below hardcodes `your-audit-logs-bucket/braintrust/*`, and the example log key is `braintrust/brainstore/...`. This enable snippet leaves `prefix` commented, so the module actually writes to `<deployment_name>/brainstore/...` and delivery fails silently.

Uncomment the prefix so the example follows the policy:

```hcl
s3_server_access_logging = {
  bucket = "your-audit-logs-bucket"
  prefix = "braintrust/"
}
```

`braintrust/*` already covers the per-bucket suffixes (`brainstore/`, `code-bundle/`, `lambda-responses/`). No extra policy statements are needed. Keep a note that the default is `<deployment_name>/` if you omit `prefix`.

> [!NOTE]
> This was found by Cursor -- took me some time to grok it.  Gist is that this is an inconsistency between the hardcoded policy and the example.

brianvans: Updated the docs

erikdw (approved):
:shipit:

PR #334, opened 2026-08-15
Pass `permissions_boundary_arn` into API ECS module

alexbasista-bt:
## Description

Threads the root module’s existing `permissions_boundary_arn` value into the `api-ecs` submodule and applies it to the API ECS task execution IAM role. It also applies the already-forwarded boundary to the Gateway ECS task execution IAM role.

Customers already setting `permissions_boundary_arn` do not need to change their configuration. The boundary will now be applied automatically to `${deployment_name}-api-ecs-task-exec` and `${deployment_name}-gateway-task-exec`, consistent with the other IAM roles created by the module.

## Context

The API ECS and Gateway ECS task execution roles were added without applying the module-wide permissions boundary. Some customers have SCPs that require a specific permissions boundary on every IAM role; without it, the roles cannot be created.

## Testing

- Ran `terraform fmt -check -recursive`
- Ran `git diff --check`
- Verified the change uses the established submodule pattern:
  - Optional string input with `default = null`
  - Root module forwards the existing value
  - IAM roles set `permissions_boundary = var.permissions_boundary_arn`
- Audited all 17 Terraform-managed IAM roles and confirmed each now applies the module-wide boundary

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -503,12 +503,13 @@ module "api_ecs" {
   alb_custom_domain              = var.braintrust_api_alb_custom_domain
   alb_drop_invalid_header_fields = var.braintrust_api_alb_drop_invalid_header_fields
 
-  kms_key_arn            = local.kms_key_arn
-  ecs_cluster_arn        = module.ecs[0].cluster_arn
-  ecs_cluster_name       = module.ecs[0].cluster_name
-  task_role_arn          = module.services_common.api_handler_role_arn
-  task_security_group_id = module.services_common.api_security_group_id
-  custom_tags            = local.all_custom_tags
+  kms_key_arn              = local.kms_key_arn
+  permissions_boundary_arn = var.permissions_boundary_arn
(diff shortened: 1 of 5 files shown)


jeffmccollum (dismissed):

brianvans:
I think there might another one to fix here https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/blob/8826df1bab4c16e44e043cc41bc0ae9b7d7c614c/modules/gateway-ecs/main.tf#L234-L240

alexbasista-bt:
> I think there might another one to fix here
> 
> https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/blob/8826df1bab4c16e44e043cc41bc0ae9b7d7c614c/modules/gateway-ecs/main.tf#L234-L240

Good 👀 @brianvans, added

brianvans (approved):

PR #336, opened 2026-08-17
Add optional Rust ingest ECS with ALB weighted % canary

erikdw:
## Summary
- Adds optional `braintrust-api-rust-ingest` ECS service + TG behind `create_rust_api_ingest` (default off).
- Two-step cutover: `enable_rust_api_ingest` points ingest ALB paths at Rust only; while create is on and enable is off, `rust_api_ingest_traffic_weight` (0–100, soak-only) can split TS/Rust.
- Ingest paths: `/logs3`, `/otel/v1/*`, `/attachment*`, `/logs3/overflow`.
- Wires Brainstore storage env/secrets for api-rs (`BRAINSTORE_INDEX_URI`, realtime WAL / locks / code-bundle / response-cache URIs, plus METADATA/WAL/REDIS secrets) — matches brainstore nodes / loadtest api-next; fixes `No index URI provided`.
- Aligned with api-rs defaults: port `8100`, health `/health/liveness`.

## Test plan
- [ ] `terraform fmt` / validate on module (+ example init if needed)
- [ ] Plan with `create_rust_api_ingest = false` → no Rust resources, ingest rules unchanged
- [ ] Plan with create true / enable false / weight `0` → Rust service/TG; ingest rules include Rust TG at weight 0
- [ ] Plan weight bump (e.g. 5) → only ALB listener rule weights change
- [ ] Plan `enable_rust_api_ingest = true` → ingest paths forward to Rust only
- [ ] Confirm task env includes `BRAINSTORE_INDEX_URI=s3://…/brainstore/index` and containers pass liveness
- [ ] Confirm image repo + tag / `VERSIONS.json` `api_rust_ingest` validation fails closed when missing

### DIFF
--- a/examples/braintrust-data-plane-external-eks-quarantine/main.tf
+++ b/examples/braintrust-data-plane-external-eks-quarantine/main.tf
@@ -151,8 +151,10 @@ module "braintrust-data-plane" {
   # Only applies when create_ai_gateway is true.
   # ai_gateway_bedrock_assume_role_arns = ["arn:aws:iam::123456789012:role/braintrust-bedrock-role"]
 
-  ### Optional Rust ingest ALB weighted canary (ECS API only; default off)
+  ### Optional Rust ingest (ECS API only; default off)
+  # Two-step: create=true, enable=false, apply; then enable=true.
   # create_rust_api_ingest                     = false
+  # enable_rust_api_ingest                     = false
   # rust_api_ingest_traffic_weight             = 0
   # rust_api_ingest_container_image_repository = "ACCOUNT.dkr.ecr.REGION.amazonaws.com/braintrust-api-rust-ingest"
   # rust_api_ingest_version_override           = "vX.Y.Z"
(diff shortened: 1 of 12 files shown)


examples/braintrust-data-plane-external-eks-quarantine/main.tf
```diff
@@ -150,4 +150,12 @@ module "braintrust-data-plane" {
   # Opt-in: bound AI Gateway Bedrock AssumeRole to approved role ARNs or IAM patterns (default: unrestricted).
   # Only applies when create_ai_gateway is true.
   # ai_gateway_bedrock_assume_role_arns = ["arn:aws:iam::123456789012:role/braintrust-bedrock-role"]
+
+  ### Optional Rust ingest (ECS API only; default off)
+  # Two-step: create=true, enable=false, apply; then enable=true.
+  # create_rust_api_ingest                     = false
+  # enable_rust_api_ingest                     = false
```
KevinGreen, line 157: We should make this language agnostic like the image name and call it something like create_api_next_ingest?

erikdw:
We aren't going with this strategy of embedding the ALB canarying into the TF dataplane module for now, instead we are using a TF overlay in each account which is a net addition on top of the existing dataplane.  This avoids either running prod-eu on a pinned branch that would drift, or adding this complicated stuff into the module proper which we definitely don't wanna do.

examples/braintrust-data-plane-external-eks-quarantine/main.tf
erikdw: Good call out, I followed that suggestion in the TF overlay strategy we are using.

PR #340, opened 2026-08-19
chore: add scorer parameter validation route

viadezo1er:
https://github.com/braintrustdata/braintrust/pull/18155 adds a route to validate the parameters of the scorers.


KevinGreen (approved):

viadezo1er:
Change of plans, there won't be a new validation endpoint, so the terraform change isn't needed.

PR #341, opened 2026-08-19
chore: add apply_immediately flag and cloudwatch_log_group outputs

kushmansingh:
## Description 

* Add a flag to skip the maintenance window for Elasticache and apply changes immediately
* Add outputs for all cloudwatch log groups

## Context 

* Tried to increase the instance size and there was no TF native way to apply it immediately
* Needed a list of log groups to forward to external monitoring

## Testing 

* TFLint


KevinGreen (approved):
Thanks for the contributions @kushmansingh!

PR #342, opened 2026-08-19
Add ElastiCache Valkey support and configurable active/passive replicas

brianvans:
> [!CAUTION]
> This is experimental functionality while PR is in draft state. Don't use in an important environment

Parameterize the ElastiCache module's engine (redis|valkey) instead of hardcoding redis. Valkey is wire-compatible, so downstream consumers (redis:// / rediss:// URL schemes, REDIS_HOST/PORT wiring) need no changes.

Also expose num_cache_clusters (1-5) on the replication group so customers can run traditional active/passive (primary + replica) with automatic failover. Defaults preserve existing behavior: engine=redis, num_cache_clusters=1, so current deployments see no diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,39 @@ If you're using a brand new AWS account for your Braintrust data plane you will 
 
 All module input variables and outputs are documented inline in the module's Terraform files (see `variables.tf`, `outputs.tf`, and the submodules for details).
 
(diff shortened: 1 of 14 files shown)


brianvans:
Actually, this is still too experimental need to test more and will re-open later when it's firmed up

PR #343, opened 2026-08-19
Harden apt against transient mirror outages in user_data

brianvans:
- Set Acquire::Retries "5" so apt retries failed .deb downloads instead of failing on the first hiccup (default is 0).
- On apt-get update failure, clear /var/lib/apt/lists/* and retry to recover from corrupt/partial index state left by a flapping mirror.


erikdw (approved):
:shipit:

KevinGreen (approved):

PR #345, opened 2026-08-20
Remove Brainstore writer restart cron

KevinGreen:
## Description 

Remove the brainstore restart cron from writers, it hasn't been necessary for a while. 

## Context 

Brainstore used to have resource contention issues that were solved by restarting.

## Testing 

We've been running several deployments without this for months.


brianvans (approved):
🫗 🫡

PR #348, opened 2026-08-21
Require a Gateway when enabling Loop runtime

erikdw:
## Summary
- Loop v2 needs Gateway semantics and cannot use the AI Proxy Lambda as the LLM router.
- `enable_loop_runtime` now fails plan unless `use_global_ai_gateway_origin` (public hosted Gateway) or `enable_ai_gateway` (private Gateway, `GATEWAY_URL` on AIProxy) is set.
- `create_ai_gateway` alone is not enough — that is the two-step cutover with AI Proxy and no Gateway. Keep `use_global_ai_gateway_origin = true` during that cutover if Loop is already on.

## Test plan
- [ ] `terraform validate` / `mise run lint` on this branch
- [ ] Confirm plan fails: `enable_loop_runtime = true`, `enable_ai_gateway = false`, `use_global_ai_gateway_origin = false` (even with `create_ai_gateway = true`)
- [ ] Confirm plan succeeds: `enable_loop_runtime = true` + `enable_ai_gateway = true`
- [ ] Confirm plan succeeds: `enable_loop_runtime = true` + `use_global_ai_gateway_origin = true`
- [ ] Examples still have Loop off and are unchanged in behavior

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -84,6 +84,17 @@ Similar two-step pattern to API ECS (`enable_ecs_api`), but gateway infra itself
 
 Use `create_ai_gateway = true` with `enable_ai_gateway = false` for a two-step prod cutover (stand up infra while keeping caller-supplied `GATEWAY_URL`, e.g. hosted gateway). Set both true for single-apply wiring on greenfield deployments.
 
+### Loop runtime requires a Gateway
+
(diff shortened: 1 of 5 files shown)


KevinGreen (approved):

erikdw:
This change is superseded by: https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/367

PR #352, opened 2026-08-25
Add Loop v2 restricted-egress PrivateLink via shared gateway sandwich

erikdw:
## Summary

Restricted Loop v2 MicroVMs live in a `10.255.0.0/16` VPC with no NAT and with DNS blocked. They cannot reach the private gateway today.

This change reuses the quarantine NLB → gateway ALB PrivateLink service. It does not add a second NLB.

- The NLB and endpoint service (`gateway_quarantine_*` names) are now the **shared provider**. Terraform creates them when quarantine PrivateLink needs them, or when restricted Loop needs them.
- The quarantine Interface VPCE is unchanged. It is created only when the quarantine flag and the module quarantine VPC are on.
- Restricted Loop gets a second Interface VPCE in the MicroVM VPC. `LOOP_RUNTIME_AI_PROXY_URL` becomes `http://<loop-vpce-dns>/v1/proxy`.
- Internet-mode Loop stays on the Function URL.
- If restricted Loop is on and this module cannot create the VPCE, apply fails unless `loop_runtime_ai_proxy_url` is set. There is no silent Function URL fallback.
- Loop v2 needs the private gateway. It cannot use AI Proxy. `use_global_ai_gateway_origin` is not a no-op for restricted Loop.

Do not test gateway-off. Loop v2 without Gateway is unsupported.

## Test plan

- [x] `terraform fmt`
- [x] `terraform validate` on the module and the production example
- [ ] Plan: internet Loop + no quarantine flag → no Loop VPCE and no extra provider
- [ ] Plan: restricted Loop + `create_ai_gateway` + module VPC → provider + Loop VPCE
- [ ] Plan: quarantine-only still creates the provider and the quarantine VPCE
- [ ] Plan: restricted Loop + existing VPC + no override → fail
- [ ] Do **not** retest with gateway off
- [ ] After draft: sandbox apply with `enable_loop_runtime = true`, `loop_runtime_sandbox_egress_mode = "restricted"`, `create_ai_gateway = true`. Confirm the Loop URL is the VPCE HTTP URL. Run one Loop turn. Confirm gateway ECS sees the traffic and the AI Proxy Function URL is not used.


erikdw:
Loop v2 should not need access to Gateway when it's using Codex with its exec-server in the MicroVMs. The Loop v2 runtime that does access Gateway won't run in the MicroVM sandboxes.

PR #355, opened 2026-08-27
Add custom CA bundle support to Brainstore

alexbasista-bt:
## Description

Extends the custom CA bundle support introduced in #350 to Brainstore.

When `custom_ca_bundle_secret_arn` is configured, Brainstore instances retrieve the PEM bundle from AWS Secrets Manager at startup (cloud-init) and pass it into the Brainstore containers through a Docker environment variable - `BRAINTRUST_CUSTOM_CA_BUNDLE`.

A separate, corresponding Brainstore application change reads this environment variable and appends the supplied certificates to its default trust store.

If the secret uses a customer-managed KMS key, `custom_ca_bundle_kms_key_arn` grants the Brainstore instance role the required scoped decrypt permission.

No Brainstore behavior changes when these inputs are unset.

> This PR is currently stacked on #350 and should be reviewed as the Brainstore portion of the custom CA bundle work.

Also, a reusable secrets retrieval function was created within the `user_data` script that includes backoff / retry logic when fetching secrets from AWS Secrets Manager (in case IAM permissions haven't propagated by the first attempt during instance startup).

## Context

Brainstore outbound HTTPS requests fail certificate verification when a customer's network uses a TLS-inspecting proxy signed by a private CA.

This extends private CA trust to Brainstore without placing the multiline PEM bundle directly in Terraform configuration or `/etc/brainstore.env`.

## Testing

- Ran `mise run lint`
- Ran `mise run validate`
- Deployed the changes to an AWS Vanilla Hybrid sandbox
- Confirmed updated Brainstore reader and writer instances launched successfully
- Confirmed both Brainstore containers started and remained healthy
- Confirmed the custom Brainstore image was running
- Confirmed Secrets Manager and customer-managed KMS access succeeded during startup
- Confirmed Brainstore outbound requests traversed the TLS-inspecting proxy without certificate or TLS errors

### DIFF
--- a/modules/brainstore-ec2/templates/user_data.sh.tpl
+++ b/modules/brainstore-ec2/templates/user_data.sh.tpl
@@ -139,6 +139,14 @@ if ! SERVICE_TOKEN_SECRET_KEY=$(aws secretsmanager get-secret-value --secret-id 
   exit 1
 fi
 
+%{ if custom_ca_bundle_secret_arn != "" ~}
+if ! BRAINTRUST_CUSTOM_CA_BUNDLE=$(aws secretsmanager get-secret-value --secret-id ${custom_ca_bundle_secret_arn} --query SecretString --output text); then
+  echo "Failed to retrieve BRAINTRUST_CUSTOM_CA_BUNDLE from Secrets Manager. Exiting with failure."
+  exit 1
+fi
+export BRAINTRUST_CUSTOM_CA_BUNDLE
+%{ endif ~}
+
 cat <<EOF > /etc/brainstore.env
 # WARNING: Do NOT use quotes around values here. They get passed as literals by docker.
 BRAINSTORE_VERBOSE=1
@@ -267,6 +275,9 @@ docker run -d \
   --network host \
   --name brainstore \
   --env-file /etc/brainstore.env \
+%{ if custom_ca_bundle_secret_arn != "" ~}
+  --env BRAINTRUST_CUSTOM_CA_BUNDLE \
+%{ endif ~}
   --restart always \
   --ulimit nofile=65535:65535 \
   -v /mnt/tmp/brainstore:/mnt/tmp/brainstore \
--- a/modules/brainstore-ec2/templates/user_data.sh.tpl
+++ b/modules/brainstore-ec2/templates/user_data.sh.tpl
@@ -130,7 +130,6 @@ systemctl enable amazon-cloudwatch-agent
 
 get_secret_value_with_retry() {
   local secret_id="$1"
-  local secret_name="$2"
   local attempt
   local retry_delay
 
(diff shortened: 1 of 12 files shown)


modules/brainstore-ec2/templates/user_data.sh.tpl
```diff
@@ -139,6 +139,14 @@ if ! SERVICE_TOKEN_SECRET_KEY=$(aws secretsmanager get-secret-value --secret-id
   exit 1
 fi
 
+%{ if custom_ca_bundle_secret_arn != "" ~}
+if ! BRAINTRUST_CUSTOM_CA_BUNDLE=$(aws secretsmanager get-secret-value --secret-id ${custom_ca_bundle_secret_arn} --query SecretString --output text); then
```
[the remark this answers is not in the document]
alexbasista-bt: Addressed with https://github.com/braintrustdata/terraform-aws-braintrust-data-plane/pull/355/commits/86a7b103549adb5595ea1f7499a4414799ca4272

modules/brainstore-ec2/templates/user_data.sh.tpl
```diff
@@ -267,6 +297,9 @@ docker run -d \
   --network host \
   --name brainstore \
   --env-file /etc/brainstore.env \
+%{ if custom_ca_bundle_secret_arn != "" ~}
+  --env BRAINTRUST_CUSTOM_CA_BUNDLE \
```
KevinGreen, line 300: This should just be echoed into the .env file instead to be consistent

modules/brainstore-ec2/templates/user_data.sh.tpl
```diff
@@ -128,16 +128,46 @@ systemctl daemon-reload
...
-  echo "Failed to retrieve SERVICE_TOKEN_SECRET_KEY from Secrets Manager. Exiting with failure."
+if ! SERVICE_TOKEN_SECRET_KEY=$(get_secret_value_with_retry "${service_token_secret_arn}" "SERVICE_TOKEN_SECRET_KEY"); then
+  exit 1
+fi
+
+%{ if custom_ca_bundle_secret_arn != "" ~}
+if ! BRAINTRUST_CUSTOM_CA_BUNDLE=$(get_secret_value_with_retry "${custom_ca_bundle_secret_arn}" "BRAINTRUST_CUSTOM_CA_BUNDLE"); then
   exit 1
 fi
+export BRAINTRUST_CUSTOM_CA_BUNDLE
```
KevinGreen, line 168: Shouldn't need to export it, the other secrets above work ok without

modules/brainstore-ec2/templates/user_data.sh.tpl
```diff
@@ -128,16 +128,46 @@ systemctl daemon-reload
...
+      return 0
+    fi
+
+    if [ "$attempt" -eq 5 ]; then
+      echo "Failed to retrieve $secret_name from Secrets Manager after $attempt attempts. Exiting with failure." >&2
+      return 1
+    fi
+
+    retry_delay=$((1 << (attempt - 1)))
+    echo "Failed to retrieve $secret_name from Secrets Manager. Retrying in $retry_delay seconds." >&2
```
KevinGreen: This could just print the arn, the name works too but saves a step in debugging if you needed to check what secret the name matches in this file, especially the db creds one because it doesn't match the env var name

modules/brainstore-ec2/templates/user_data.sh.tpl
KevinGreen: I guess the reasoning is that it's a secret so keeping it out of a file helps avoid it being readable but if someone is on the host they can read the env vars from the process or Docker with the same as the file

modules/brainstore-ec2/templates/user_data.sh.tpl
alexbasista-bt: Updated to print the secret ARN

modules/brainstore-ec2/templates/user_data.sh.tpl
alexbasista-bt: The CA bundle is multiline PEM but Docker’s `--env-file` format is line-oriented, so echoing it directly into `brainstore.env` would corrupt the value (with how it handles new lines). Passing the exported variable with `--env BRAINTRUST_CUSTOM_CA_BUNDLE` preserves the newlines

modules/brainstore-ec2/templates/user_data.sh.tpl
alexbasista-bt: After chatting we decided to keep the `export` because Docker’s name-only `--env BRAINTRUST_CUSTOM_CA_BUNDLE` reads from the inherited process environment, while passing the multiline PEM explicitly would expose it in the command arguments

KevinGreen (approved):

PR #359, opened 2026-08-29
Route ECS API self-calls through the internal ALB

brianvans:
## Description

 Routes ECS API self-calls through the internal API ALB:

 - Sets AI_PROXY_FN_URL, TS_API_ASYNC_SCORING_PROXY_URL, and BRAINTRUST_API_URL to local.api_ecs_url.
 - Leaves PROXY_URL on loopback for the local proxy path.
 - Does not change Lambda, gateway, or loop-runtime wiring.

 No customer configuration changes required.

## Context

 ECS API tasks run with IS_LOCAL=true. Async scoring writebacks and invoke dispatch previously used loopback URLs, bypassing ALB path-based routing and concentrating PostgreSQL/TLS work on the calling task.

 Using the internal ALB URL routes those calls through the appropriate ingest/background pools.

### DIFF
--- a/modules/api-ecs/locals.tf
+++ b/modules/api-ecs/locals.tf
@@ -53,6 +53,7 @@ locals {
     PRIMARY_ORG_NAME                                  = var.primary_org_name
     ALLOWED_ORG_IDS                                   = var.allowed_org_ids
     BRAINTRUST_DEPLOYMENT_NAME                        = var.deployment_name
+    BRAINTRUST_API_URL                                = local.api_ecs_url
     BRAINTRUST_LEGACY_IDS                             = "true" # emit v3 spans in py+ts sdks. remove this setting when v4 migration is complete
     RESPONSE_BUCKET                                   = var.response_bucket
     CODE_BUNDLE_BUCKET                                = var.code_bundle_bucket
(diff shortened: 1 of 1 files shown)


modules/api-ecs/locals.tf
```diff
@@ -77,7 +78,7 @@ locals {
     TELEMETRY_LOG_LEVEL                               = var.billing_telemetry_log_level
     INSERT_LOGS2                                      = "true"
     NODE_MEMORY_PERCENT                               = "80"
-    AI_PROXY_FN_URL                                   = "http://127.0.0.1:8000"
+    AI_PROXY_FN_URL                                   = local.api_ecs_url
```
erikdw, line 81: This change is a no-op, because the ECS code doesn't use it. Suggest dropping it.

> [!note]
> Cursor's explanation for why this is a no-op: 
> 
> `AI_PROXY_FN_URL` is only read in the Lambda invoke helpers (`IS_LOCAL=false`), where the destination is `AI_PROXY_FN_ARN` and the URL hostname is used later to rebuild `/v1/proxy` for scorers. 
>
> ECS has `IS_LOCAL=true`, so that code never runs.

brianvans:
Closing while we look into alternatives

PR #364, opened 2026-09-02
Add first-party DuckDB Node API Lambda layer

alsmola:
## Summary

- create a deployment-owned DuckDB Node API Lambda layer from the versioned regional asset bucket
- use the effective Lambda release tag for both Lambda code and the DuckDB layer
- attach the first-party layer to APIHandler, AIProxy, AutomationCron, and QuarantineWarmupFunction
- remove the old public DuckDB layer dependency
- fail planning when the selected Lambda release has no matching DuckDB layer pointer

This is the Terraform counterpart to [braintrust#19491](https://github.com/braintrustdata/braintrust/pull/19491). The required publisher support landed in [braintrust#19867](https://github.com/braintrustdata/braintrust/pull/19867).

## Rollout safety

The v2.13.0 DuckDB layer pointer and referenced artifact are available in all 12 regional asset buckets. The module now uses lambda_version_tag_override, when set, for both Lambda code and the layer; otherwise both use the locked v2.13.0 release.

A pre-v2.13.0 override has no matching first-party layer pointer and intentionally fails during planning rather than pairing incompatible Lambda code and layer dependencies.

## Testing

- terraform fmt -check -recursive
- terraform validate
- tflint --recursive
- terraform test: 20 passed, 0 failed
- dedicated plan coverage for a Lambda version override with its matching first-party layer


modules/services/main.tf
```diff
@@ -44,9 +45,15 @@ locals {
   code_bundle_bucket_id      = split(":::", var.code_bundle_bucket_arn)[1]
   lambda_responses_bucket_id = split(":::", var.lambda_responses_bucket_arn)[1]
 
-  # Lambda versions can be specified statically through VERSIONS.json or dynamically via lambda_version_tag_override
-  # If lambda_version_tag_override is provided, use it. Otherwise, use the lambda_version_tag from VERSIONS.json
-  lambda_version_tag = var.lambda_version_tag_override != null ? var.lambda_version_tag_override : jsondecode(file("${path.module}/VERSIONS.json"))["lambda_version_tag"]
+  # Lambda versions can be specified statically through VERSIONS.json or dynamically via overrides.
+  locked_versions    = jsondecode(file("${path.module}/VERSIONS.json"))
+  lambda_version_tag = var.lambda_version_tag_override != null ? var.lambda_version_tag_override : local.locked_versions["lambda_version_tag"]
+  duckdb_node_api_layer_version_tag = var.lambda_version_tag_override != null ? (
```
KevinGreen: Since the layer is versioned with the releases I'm not sure we need to have this control to override it

alsmola: From Codex:
> You’re right; we can use the Lambda override for both. I separated them because EU currently pins an older pre-layer SHA. We can remove the separate control if we bump EU to v2.13.0 at the same time. The only consideration is that its current shared version override also rolls Brainstore, API ECS, and Loop Runtime, so I’ll coordinate that as a production infra bump.

alsmola: Agreed. Updated in f567d5c: the DuckDB layer now uses the effective Lambda version tag directly, so lambda_version_tag_override selects both artifacts. I removed the separate layer override and legacy fallback. A release without a matching layer pointer now fails during planning, which prevents an incompatible Lambda/layer pairing.

KevinGreen (approved):

PR #365, opened 2026-09-02
Always enable Loop runtime metrics and traces

ankrgyl:
## Summary

- append metrics and traces to the shared control-plane telemetry configuration for Loop runtime
- preserve all deployment-configured telemetry signals

## Testing

- terraform fmt -check modules/loop-runtime-ecs/main.tf
- terraform validate
- terraform test (19 passed)

### DIFF
--- a/modules/loop-runtime-ecs/main.tf
+++ b/modules/loop-runtime-ecs/main.tf
@@ -48,7 +48,7 @@ locals {
       BRAINSTORE_REALTIME_WAL_URI        = "s3://${local.brainstore_s3_bucket}/brainstore/wal"
       BRAINSTORE_CODE_BUNDLE_URI         = "s3://${var.code_bundle_bucket}"
       BRAINSTORE_CACHE_DIR               = "/mnt/tmp/brainstore"
-      BRAINSTORE_CONTROL_PLANE_TELEMETRY = var.monitoring_telemetry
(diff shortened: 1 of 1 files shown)


modules/loop-runtime-ecs/main.tf
```diff
@@ -48,7 +48,7 @@ locals {
       BRAINSTORE_REALTIME_WAL_URI        = "s3://${local.brainstore_s3_bucket}/brainstore/wal"
       BRAINSTORE_CODE_BUNDLE_URI         = "s3://${var.code_bundle_bucket}"
       BRAINSTORE_CACHE_DIR               = "/mnt/tmp/brainstore"
-      BRAINSTORE_CONTROL_PLANE_TELEMETRY = var.monitoring_telemetry
+      BRAINSTORE_CONTROL_PLANE_TELEMETRY = join(",", [var.monitoring_telemetry, "metrics", "traces"])
```
brianvans, line 51:
```suggestion
      BRAINSTORE_CONTROL_PLANE_TELEMETRY = join(",", sort(tolist(setunion(
        toset(compact(split(",", var.monitoring_telemetry))),
        toset(["metrics", "traces"]),
      ))))
```

If the consumer just does a map+contains we might not care about dupes, but if we do, we can do a setunion to remove them

brianvans (approved):

modules/loop-runtime-ecs/main.tf
ankrgyl: we dont care about dups

PR #367, opened 2026-09-03
Point Loop Runtime model traffic at the private gateway

erikdw:
## Summary
- `LOOP_RUNTIME_AI_PROXY_URL` is used by Loop runtime, not by exec-server in the MicroVM. Stop sending that in-VPC caller through the AI Proxy Function URL.
- When `create_ai_gateway` is set, Loop now calls the private gateway ALB `/v1/proxy`. Gateway ALB ingress is opened from the Loop task SG (separate rule, no cycle with `authorized_security_groups`).
- Without a private gateway, Loop falls back to the hosted gateway origin or CloudFront API `/v1/proxy` — still not the Function URL.

## Test plan
- [x] CI: Terraform Lint & Validate
- [x] Applied on erikdw-sandbox (Loop + private gateway already present): `LOOP_RUNTIME_AI_PROXY_URL` is `http://<gateway-alb>/v1/proxy`, new ALB :80 ingress from the Loop task SG, Reachability Analyzer Loop ENI → gateway ALB reachable
- [x] Live Loop turn via CloudFront `/loop/runtime` (`gpt-5-nano` on `braintest`): Loop `base_url` is the private gateway ALB, turn completed with model usage (no Function URL)
- [x] `enable_loop_runtime=true` → `false`: Loop ECS + `gateway_from_task` destroyed; gateway ALB ARN unchanged and still serving. CloudFront Loop VPC origin delete 409 in the same apply (origin still attached); restored Loop afterward
- [x] `create_ai_gateway=true` → `false`: same apply set `loop_runtime_ai_proxy_url` to `https://gateway.braintrust.dev/v1/proxy` and destroyed `gateway_from_task`. CloudFront PrivateGatewayOrigin delete 409 in the same apply (same async detach). Restored `create_ai_gateway`; Loop env, SG rule, and gateway listener are back


loop-runtime.tf
```diff
@@ -15,6 +15,25 @@ locals {
...
+    "https://%s/v1/proxy",
+    trimsuffix(replace(var.global_ai_gateway_origin_domain, "/^https?:\\/\\//", ""), "/"),
+  )
+  loop_runtime_ai_proxy_url = (
+    local.create_ai_gateway
+    ? "${one(module.gateway_alb[*].gateway_url)}/v1/proxy"
+    : (
+      var.use_global_ai_gateway_origin
+      ? local.hosted_ai_gateway_proxy_url
+      : one(module.ingress[*].api_url)
```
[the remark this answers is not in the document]
erikdw: The env var on this fallback is the CloudFront root, which is intentional. Loop does not request that URL as-is — `normalized_model_proxy_url` appends `/v1/proxy` unless the path is already there (`brainstore/agent_runtime/src/cli.rs`), and that's the CloudFront behavior that selects the model-proxy origin. Same host-only contract as the old Function URL / ECS ALB values. Private-gateway and hosted-origin stacks never take this branch. Appending `/v1/proxy` in Terraform would only make env dumps look consistent; it isn't needed for routing.

loop-runtime.tf
```diff
@@ -15,6 +15,24 @@ locals {
...
+  # Loop ECS task model proxy. Prefer the in-VPC private gateway ALB
+  # when it exists; otherwise hosted gateway or CloudFront API /v1/proxy.
+  # Do not use the AI Proxy Function URL — that hop is for internet-mode /
+  # quarantine callers, not Loop.
+  hosted_ai_gateway_proxy_url = format(
+    "https://%s/v1/proxy",
+    trimsuffix(replace(var.global_ai_gateway_origin_domain, "/^https?:\\/\\//", ""), "/"),
+  )
+  loop_runtime_ai_proxy_url = (
+    local.create_ai_gateway
```
soldatchenko: Separate from the `/v1/proxy` normalization discussed below: could this cut Loop over before the staged gateway rollout is ready?

With `create_ai_gateway=true` and `enable_ai_gateway=false`, this immediately points Loop at the new ALB. Since Loop depends on `gateway_alb`, not `gateway_ecs`, could its rollout finish before the gateway is healthy—or remain pointed there if gateway deployment fails?

Should this switch on `enable_ai_gateway`, or otherwise wait for gateway health?

soldatchenko (changes requested):

soldatchenko (dismissed):

loop-runtime.tf
erikdw: Good catch!  Yes, I'll push a fix to gate on `enable_ai_gateway` instead.

KevinGreen (approved):

PR #376, opened 2026-09-10
BYOC management role self deny

alexbasista-bt:
## Description
Adds an explicit deny preventing the BYOC management role from directly modifying or deleting itself, or being passed to an AWS service. The setup script inserts the role’s exact ARN into the policy, including when a custom role name is used.

Also:
- Handles both string and object trust policy responses when preserving the External ID
- Reduces the maximum role session duration from four hours to one hour to match AWS STS’s one-hour limit for role chaining

Existing installations can apply these changes by rerunning the setup script. If the BYOC management role uses a custom name (rather than the default `BraintrustManagementRole`), specify it with `--role-name`.

## Context
Adds a targeted safeguard around the management role while preserving its ability to manage infrastructure roles and their policies.

## Testing 
- Ran `create-management-role.sh` and verified templating syntax - BYOC management role created successfully.
- Completed a Terraform apply - successful data plane deployment.
- Successfully ran a Terraform destroy.
- Confirmed the assumed management role identity through CloudTrail.
- IAM policy simulation returned an explicit deny for `iam:PutRolePolicy` on the management role.
- Shell syntax, template rendering checks, and `mise run lint` passed.

### DIFF
--- a/managed-byoc/create-management-role.sh
+++ b/managed-byoc/create-management-role.sh
@@ -56,7 +56,7 @@ resolve_external_id() {
     existing="$(aws iam get-role \
       --profile "$PROFILE" \
       --role-name "$ROLE_NAME" \
-      --output json | jq -r '.Role.AssumeRolePolicyDocument | fromjson | .Statement[0].Condition.StringEquals."sts:ExternalId" // empty')"
+      --output json | jq -r '.Role.AssumeRolePolicyDocument | (if type == "string" then fromjson else . end) | .Statement[0].Condition.StringEquals."sts:ExternalId" // empty')"
     if [[ -n "$existing" && "$existing" != "null" ]]; then
       EXTERNAL_ID="$existing"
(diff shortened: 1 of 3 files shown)


managed-byoc/create-management-role.sh
```diff
@@ -139,32 +159,40 @@ if aws iam get-role --profile "$PROFILE" --role-name "$ROLE_NAME" >/dev/null 2>&
     --role-name "$ROLE_NAME" \
     --policy-document "$TRUST_POLICY_DOCUMENT" \
     >/dev/null
-  echo "Updating max session duration to 4 hours..."
+  echo "Updating max session duration to 1 hour..."
```
KevinGreen: It should be really rare but in case of some rare event like ec2 instance availability I've seen deployments (elsewhere) take over an hour. Could we keep this at 4 hours for now?

PR #386, opened 2026-09-16
Add optional sandbox observability stack

brain-eugenevignanker:
## Summary
- add optional Grafana/Loki/Prometheus/Tempo/OpenTelemetry EC2 stack to the sandbox example
- package and upload the local observability compose bundle for bootstrap
- wire Brainstore OTLP telemetry to a stable private Route 53 DNS endpoint when enabled

## Validation
- terraform init -backend=false
- terraform validate


KevinGreen (changes requested):
@brain-eugenevignanker We should make a separate module for this in a different repo instead

